PR #71, opened 2026-03-16
feat: show power consumption in power menu

### LINKED ISSUE #23: Show power consumption in power menu

4RH1T3CT0R7:
- Display INA219 voltage, current, and power readings in the power menu header
- Values update every 1 second via periodic FuriEventLoopTimer
- Uses existing Power service API (power_ina219_get_voltage_v/current_a/power_w)

Closes #23

### DIFF
--- a/applications/services/desktop/desktop.c
+++ b/applications/services/desktop/desktop.c
@@ -212,6 +212,9 @@ static void desktop_scene_event_logic(FuriEventLoopObject* object, void* context
         scene_enter(desktop->debug_menu_scene, desktop);
         consumed = true;
         break;
+    case DesktopSceneEventTypePowerUpdate:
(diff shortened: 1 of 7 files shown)


DrZlo13 (approved):

DrZlo13 (changes requested):

4RH1T3CT0R7:
The power menu moved into the desktop scenes since I opened this, so rebasing turned into a reimplementation on the new PopupMenu: optional status line under the title, with the border under it.

No allocation now, the model formats into a fixed `char[32]`.

It refreshes on the 5s power event rather than a 1s timer, since scenes can't reach the event loop. Each refresh is three blocking round-trips to the power thread; `header_update_battery_state` is commented out, maybe for the same reason. Happy to cache in the power service instead.

PR #129, opened 2026-05-21
Power: OTG enable/disable and voltage/current control

DrZlo13:
# What's new

- [ Describe changes here ]

# Verification 

# Contributor checklist


applications/services/power/power.c
```diff
@@ -182,6 +194,80 @@ API_WRAPPER(bq25792_watchdog_reset, Bq25792Status, Bq25792*);
...
+        // keep WD enabled until OTG is confirmed off.
+        res = bq25792_otg_enable(instance->bq25792_header, false);
+        if(res != Bq25792StatusOk) {
+            FURI_LOG_E(TAG, "Failed to disable OTG: %d", res);
+            furi_event_loop_timer_stop(instance->otg_watchdog_timer);
+            return res;
+        }
+
+        instance->otg_enabled = false;
+        furi_event_loop_timer_stop(instance->otg_watchdog_timer);
```
Skorpionm: The disabling of `otg_watchdog_timer` occurs independently of the execution results of `bq25792_otg_enable`; therefore, the call to `furi_event_loop_timer_stop(instance->otg_watchdog_timer);` can be moved to a position before the call to `bq25792_otg_enable`.

applications/services/power/power.c
```diff
@@ -79,8 +86,16 @@ static void power_bq25792_print_charger_irq(Power* instance) {
...
-    FURI_LOG_I(TAG, "  IRQ0:    0x%02X %s", fl.data[0], furi_string_get_cstr(arena));
+    if(furi_string_size(arena) != 0) FURI_LOG_I(TAG, "  IRQ0:    0x%02X %s", fl.data[0], furi_string_get_cstr(arena));
+
+    if(fl.flag0.iindpm_flag && instance->otg_enabled) {
+        FURI_LOG_W(TAG, "OTG overcurrent (IINDPM/IOTG) detected");
+        FURI_CRITICAL_ENTER();
+        PowerBq25792OtgOvercurrentCallback cb = instance->otg_overcurrent_callback;
+        void* cb_ctx = instance->otg_overcurrent_context;
+        FURI_CRITICAL_EXIT();
+        if(cb) cb(cb_ctx);
```
Skorpionm: It is not strictly necessary to declare variables for assigning `cb` and `cb_ctx` by wrapping them within a critical section; instead, you can invoke the callback directly, thereby saving CPU time by eliminating the overhead of processing the critical section.

if(instance->otg_overcurrent_callback) instance->otg_overcurrent_callback(instance->otg_overcurrent_context);

Skorpionm: Additionally, the OTG interrupt handling logic should be extracted from the function responsible for displaying triggered interrupt information and moved into a separate, dedicated function.
This is because the output of such information is not always required, and the process itself takes a "very long time"—specifically, the time needed to format the output strings and transmit them to the debug console.

applications/services/power/power.c
```diff
@@ -182,6 +194,80 @@ API_WRAPPER(bq25792_watchdog_reset, Bq25792Status, Bq25792*);
...
+        if(res != Bq25792StatusOk) {
+            FURI_LOG_E(TAG, "Failed to set OTG watchdog time: %d", res);
+            return res;
+        }
+        res = bq25792_watchdog_reset(instance->bq25792_header);
+        if(res != Bq25792StatusOk) {
+            FURI_LOG_E(TAG, "Failed to reset OTG watchdog: %d", res);
+            return res;
+        }
+        res = bq25792_otg_enable(instance->bq25792_header, true);
```
Skorpionm, line 241: Implement a mechanism to disable the watchdog if `bq25792_otg_enable` fails.

Skorpionm: However, the `otg_watchdog_timer` will not start if something goes wrong, and the bq25792 will automatically restart itself.

PR #131, opened 2026-05-23
Add GPL-3.0 license

### LINKED ISSUE #122: Add LICENSE to repo
Please add license

descambiado:
Adds a GPL-3.0 license file to the repository.

The project currently has no license file, which means contributors and users have no clarity on the terms under which the code can be used, modified, or distributed. GPL-3.0 matches the license used across other Flipper Devices firmware repositories.

Closes #122

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,675 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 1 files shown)


DrZlo13:
Thanks for your contribution, but we are currently discussing within the team which license we should take, because GPL3 seems too restrictive.

descambiado:
Understood, thanks for the heads-up. Happy to wait for the team's decision on licensing direction. If you settle on a more permissive license (Apache 2.0, MIT, BSD, etc.) I can rebase this PR to use that instead, or close it in favor of a commit from the team. Either path works for me.

PR #134, opened 2026-05-23
Power: USB port state - Sink/source, overcurrent

### LINKED ISSUE #40: Power: USB port state
Sink/source, overcurrent.

basarbyz:
Closes #40
# What's new
- FUSB302 driver: new `fusb302_get_port_state()` reads the `TOGSS` bitfield from `STATUS1A` and maps it to a `Fusb302PortState` enum (source/sink on CC1/CC2, audio accessory, or still toggling).
- FUSB302 driver: `fusb302_read_role()` now returns `Fusb302ReadRoleResult` (`Nothing` / `ToggleDone` / `Overcurrent`) instead of `bool`, so faults are distinguishable from no-ops.
- FUSB302 driver: unmasked the `I_OCP_TEMP` interrupt (`m_ocp_temp = 0`) to enable VCONN overcurrent / overtemperature reporting.
- PD service: added `PdPortState` enum and `PdEvent` struct to `pd.h` as a driver-agnostic public API, including `PdPortStateOvercurrent` (sourced only from the interrupt path, never from TOGSS).
- PD service: rewrote the ISR callback (previously a commented-out polling stub). On `ToggleDone` it reads port state and publishes a `PdEvent`; on `Overcurrent` it publishes an overcurrent event. Early-returns when `PdModeOff`.
- Events are delivered via the existing PD `furi_pubsub`, so subscribers can react without polling or coupling to PD internals.

# Verification
- Build the firmware and flash to the device.
- Plug in a USB-C source with the cable in one orientation; confirm a `PdEvent` is published with the expected sink-on-CC1 or sink-on-CC2 state. Flip the cable and confirm the opposite CC line is reported.
- Plug into a sink (e.g. another device acting as a consumer); confirm the source-on-CC1 / source-on-CC2 states are reported correctly.
- Plug in an audio accessory adapter; confirm the audio accessory state is reported.
- Trigger a VCONN overcurrent (e.g. short VCONN to GND through a current-limited path); confirm a `PdPortStateOvercurrent` event is published and that it arrives via the interrupt path.
- Set PD mode to off and repeat a plug event; confirm no events are published (early return path).
- Subscribe to the PD pubsub from a simple test consumer and confirm exactly one event fires per state change.

**Note:** The firmware builds successfully locally. I don't have hardware for on-device verification - someone with hardware testing against the steps above is requested! 

# Contributor checklist

- [ ] I have tested these changes myself (on hardware or in simulation)

### DIFF
--- a/applications/services/pd/pd.c
+++ b/applications/services/pd/pd.c
@@ -85,31 +85,39 @@ static void pd_custom_event_callback(uint32_t events, void* context) {
     Pd* instance = (Pd*)context;
 
     if(events & PdEventTypeIsr) {
-        // fusb302_read_role(instance->fusb302_header);
+        if(instance->mode == PdModeOff) {
+            return;
+        }
 
-        // fusb302_pd_reset_hard(instance->fusb302_header);
+        Fusb302ReadRoleResult role_result = fusb302_read_role(instance->fusb302_header);
 
-        // fusb302_cc_orientation_set(instance->fusb302_header, PdTypeCcOrientationNormal);
-        // fusb302_pd_reset_logic(instance->fusb302_header);
-        // fusb302_pd_autogoodcrc_set(instance->fusb302_header, true);
-        // fusb302_pd_autoretry_set(instance->fusb302_header, 3);
-        // fusb302_pd_rx_flush(instance->fusb302_header);
+        if(role_result == Fusb302ReadRoleResultToggleDone) {
+            Fusb302PortState fusb302_port_state = Fusb302PortStateUndefined;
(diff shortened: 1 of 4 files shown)


DrZlo13:
Good afternoon.

Unfortunately this PR is not usable in its current form. While the CPU is running, the MCU has no access to the FUSB302 or the I2C bus in general, including the current sensors and other devices on it.

We are working on fixing this in the zlo/usb-controller branch, which will include a full FUSB302 driver along with the USB-PD stack.

I'm closing this PR for now. I also want to apologize that the current state of the issue tracker doesn't reflect the full picture of problems and their dependencies. We will address that.

basarbyz:
Got it, thanks. I'll keep an eye on the help-wanted tasks.

PR #137, opened 2026-05-26
fix(pio_i2c): implement timeout on busy-wait loops to prevent system hang

electricalgorithm:
## Changeset
```
The `until` parameter in `pio_i2c_write_blocking` and `pio_i2c_read_blocking` was documented 
with `//Todo: implement timeout` but was entirely unused (`UNUSED(until)`). All internal busy-
wait loops (`pio_i2c_put16`, `pio_i2c_put_or_err`, `pio_i2c_wait_idle`) spun indefinitely with 
no timeout. If the PIO state machine gets stuck (e.g., slave holds SCL low, bus contention), 
the CPU hangs permanently.

The fix:
- Thread the `absolute_time_t until` timeout through all helper functions: `pio_i2c_put16`, 
`pio_i2c_put_or_err`, `pio_i2c_wait_idle`, `pio_i2c_start`, `pio_i2c_stop`, `pio_i2c_repstart`
- Each busy-wait loop now checks `time_us_64() >= until` and returns early on timeout, 
propagating `PICO_ERROR_GENERIC` to the caller
- `pio_i2c_put16` and `pio_i2c_put_or_err` now return `bool` so callers can detect timeout 
failures
- Removed `UNUSED(until)` and the TODO comment

Build: cmake build succeeds with no warnings.
Test: Flash the firmware to Pico 2, and run I2C CLI tool. Since it's hardware-dependent timeout 
fix, it is a hard to reproduce without proper tools.
```

## What's New

These loops never give up, there's no deadline condition (a timeout, an error, or so):

```c
// pio_i2c_put16 — address byte, read fill bytes
while(pio_sm_is_tx_fifo_full(instance->pio, instance->sm))  // no error check, no timeout
    ;

// pio_i2c_wait_idle — waits for TX to drain
while(!(txstall || pio_i2c_check_error(instance)))  // error check but no timeout
    tight_loop_contents();
```

If a slave holds SCL low, the TX FIFO never drains, and the CPU spins forever. The `until` parameter existed in the API but was explicitly discarded:

```c
int pio_i2c_write_blocking(..., absolute_time_t until) {
    //Todo: implement timeout
    UNUSED(until);
    pio_i2c_put16(instance, ...);
    pio_i2c_wait_idle(instance);
```

The fix I applied is to use `until` through the call chain. Each loop now shares the until variable through the chain:

```c
static inline bool pio_i2c_put16(I2cMasterPio* instance, uint16_t data, absolute_time_t until) {
    while(pio_sm_is_tx_fifo_full(instance->pio, instance->sm)) {
        if(until != 0 && time_us_64() >= until) return false;
    }
    ...
    return true;
}

static void pio_i2c_wait_idle(I2cMasterPio* instance, absolute_time_t until) {
    while(!(txstall || pio_i2c_check_error(instance))) {
        if(until != 0 && time_us_64() >= until) break;
        tight_loop_contents();
    }
}
```

I also changed the function signature `pio_i2c_put16` and `pio_i2c_put_or_err` to return a bool. It is not directly required, but I believe better for error interpretation for consumers. 

## Verification 

It's hard to reproduce the issue without proper hardware. Therefore, I could only prepare a CLI debug tool to reproduce the issue. The CLI command needed to have a patch on the I2C driver code, but I believe it does not affect anything regarding to functionality. Find the CLI reproducer implementation patch in the following file; I'm not adding it to the tree:

[pio_i2c_timeout_test_cli_reproducer.patch](https://github.com/user-attachments/files/28260256/pio_i2c_timeout_test_cli_reproducer.patch)

1. Claims a PIO block first, loads the I2C program, starts the SM, and configures GPIOs for PIO.
2. Calls `pio_sm_set_enabled(false)` to freeze the SM so it can't consume anything, then stuffs `pio_sm_put()` until pio_sm_is_tx_fifo_full()` returns true. SM stays paused.
3. Then, calls `pio_i2c_write_blocking(i2c, 0x42, {0xAA, 0xBB}, 2, false, now + 10ms)`
4. It calls `pio_i2c_start()` which calls `pio_i2c_put_or_err()` to push the START sequence.
5. First `put_or_err` call: `pio_sm_is_tx_fifo_full()` is true (since we filled it in step 2, SM is paused).
6. Now it enters spin-loop.

The test case doesn't reimplement the timeout directly, but fills the FIFO of the SM to reproduce something similar. I first ran it with the bad code, and then, after the patch (good code), I introduced it here.

### Bad Code (Before Patch)
```
Firmware version: dev r93 (ad12260-dirty built on 2026-05-26)

>: ?
Commands available:
?                             led
clock_out                     log
expander_ext                  otp
free                          pio_i2c_timeout_test
free_blocks                   power
gpio                          power_consumption
help                          top
i2c                           uptime
>: pio
pio_i2c_timeout_test

>: pio_i2c_timeout_test
PIO I2C timeout test
pio_i2c_init OK
TX FIFO filled
```

After the last log, the hardware is frozen, and I had to re-plug the device. I don't know why watchdog is not fired. 

### Good Code (After Patch)
```
Firmware version: feat/add-timeout-to-pio-i2c r94 (6d7cb3d-dirty built on 2026-05-26)

>: ?
Commands available:
?                             led
clock_out                     log
expander_ext                  otp
free                          pio_i2c_timeout_test
free_blocks                   power
gpio                          power_consumption
help                          top
i2c                           uptime
>: pio
pio_i2c_timeout_test

>: pio_i2c_timeout_test
PIO I2C timeout test
pio_i2c_init OK
TX FIFO filled
write_blocking returned -1 after 10059 us
PASS: write timed out as expected
```

## Contributor Checklist


electricalgorithm:
Can you review it, please? cc: @DrZlo13 @Skorpionm @zhovner

DrZlo13:
Thanks for the contribution! We don't have a license in the project yet. I'm waiting on a decision about which one to go with. I think having a license in place before accepting contributions is important, so I appreciate your patience.

We're currently leaning towards a permissive license like MIT or BSD. Would you be okay with that?

electricalgorithm:
> Thanks for the contribution! We don't have a license in the project yet. I'm waiting on a decision about which one to go with. I think having a license in place before accepting contributions is important, so I appreciate your patience.

I agree. Thanks for the information. I was quite curious on the future of the project as well.

> We're currently leaning towards a permissive license like MIT or BSD. Would you be okay with that?

All fine by me. I just want to contribute to this nice project, since I like the idea behind it. Doing my best to support you folks.

Skorpionm:
Have you actually tried to reproduce the problem? On real hardware using this driver, or is this just a theoretical conclusion? When I added this "todo," I also thought theoretically that this could happen, but when testing on real hardware, physically pulling the SDA and SCL lines to 0, I didn't observe any looping. The driver restarted transactions.

DrZlo13 (approved):

PR #138, opened 2026-05-26
fix(i2c_intercom): add volatile qualifier to ISR-shared state fields

electricalgorithm:
## What's New

The `state` and `mem_address` fields in `I2CIntercom` are accessed by two actors: I2C ISR (`i2c_intercom_isr`) and the alarm timeout callback (`i2c_intercom_timeout_callback`). On a dual-core CPU these can run concurrently on different cores. Without `volatile` keyword, the compiler may cache these values in registers, causing one execution context to read stale state after the other has updated it.

Struct Definition:
```c
struct I2CIntercom {
    ...
    I2CIntercomState state;
    size_t mem_address;
};
```

Interrupt routine accessing the state:
```c
instance->state = I2CIntercomStateStart;
```

Alarm callback accessing the state:
```c
if(instance->state == I2CIntercomStateStart) {
}
```

Therefore, I added the `volatile` keyword to `state` and `mem_address`, and cast it correctly when passing `&instance->mem_address` to `i2c_intercom_receive_address`.

## Verification
It's a really hard race condition to reproduce. Furthermore, I do not have the proper toolset to generate it as well. The only verification I did is to build the firmware and flash it to Pico 2 and check if the shell is up via USB.

# Contributor Checklist


electricalgorithm:
Somehow, I cannot ask for reviewers. Therefore, I'm adding you guys. Sorry.

cc: @DrZlo13 @Skorpionm @zhovner

DrZlo13 (approved):
Good catch!

PR #144, opened 2026-05-28
Licensing

### LINKED ISSUE #122: Add LICENSE to repo

DrZlo13:
# What's new

- Added top-level REUSE.toml describing licensing for the whole repo: `MIT` by default, `GPL-3.0-or-later` for the iles ported from flipperzero-firmware, `CC-BY-SA-4.0` for assets/, docs/ and *.md, plus correct third-party tags. Submodules are left untouched.
- Closes #122

# Verification 

- Run `reuse download MIT GPL-3.0-or-later CC-BY-SA-4.0 BSD-3-Clause Zlib` then `reuse lint` — should report the project as REUSE-compliant.

# Contributor checklist

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,11 @@
+# Flipper One MCU firmware license information
+
+Source code in this repository is covered by several licenses.
+
+Most of the first-party code is licensed under MIT. A small subset of files ported from the [Flipper Zero firmware](https://github.com/flipperdevices/flipperzero-firmware) is GPL-3.0-or-later.
+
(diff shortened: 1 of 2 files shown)


DrZlo13:
### The switch from our usual GPL to MIT

At some point we realized that Furi is quite useful on its own and we'd like to see it adopted in more products. To enable that, we relicensed Furi under MIT — and as a continuation, I now want to put the entire Furi-to-RP2350 glue layer under MIT as well.

Under GPL I've kept only the files that:
- originate from Flipper Zero and have commits in their history from people outside our development team;
- are application code that would be useless outside of this project.

gsurkov (approved):

Issue #146, opened 2026-05-29
CI: Tracking memory footprint with MemBrowse


michael-membrowse:
Hey everyone,

I've been working on a tool called [MemBrowse](https://membrowse.com/) that tracks firmware memory footprint across commits,
and I thought it might be useful for Flipper One firmware that it's meant to run on a small MCU.

To demonstrate how it works, I set up a public dashboard with history builds:
[https://membrowse.com/public/michael-membrowse/flipperone-mcu-firmware](https://membrowse.com/public/michael-membrowse/flipperone-mcu-firmware)

What it does:

1. Runs automatically on pull request updates and on merges to the main branch
2. Tracks FLASH and RAM usage per commit with historical trends
3. Creates a PR comment with detailed memory footprint changes for all tracked targets
4. Shows diffs between any two commits so you can see exactly what changed
5. Breaks down memory usage by region, section, library and symbol
6. Supports feature budgets using thresholds defined via commit message keywords

Integration:

It runs as a GitHub Action, parses ELF files and linker scripts and generates a report that is uploaded to membrowse.com.
No code changes are required in the repo (besides the github action), and very little ongoing maintenance once set up.
If there’s interest, I’d be happy to help set up and maintain it when needed.

Would such memory tracking be useful for the Flipper One firmware development?

Skorpionm:
It no longer shows RAM correctly. You're probably only counting the amount of RAM allocated by static variables.

michael-membrowse:
Hi @Skorpionm,

The RAM footprint contains the functions and global variables that are set to be in the RAM.
What symbols do you think are missing?

<img width="1554" height="1073" alt="Image" src="https://github.com/user-attachments/assets/85604773-7d04-4068-8f14-83d15a6621ee" />

michael-membrowse:
Its currently just a snapshot. To continuously monitor changes, it will need to be in flipperone-mcu-firmware CI.
I can offer a PR for that integration and help maintaining it.

DrZlo13:
Hello,

While working on Flipper Zero, we developed a [similar tool](https://fw-reports.flipp.dev/) in-house, but we'd be interested in trying out your solution. As I understand it, it's free for open-source projects?

michael-membrowse:
Hi, yes, it is free for open source. I will create a PR then.

michael-membrowse:
Please see #148

Skorpionm:
[...]

PR #148, opened 2026-05-29
ci: membrowse integration

michael-membrowse:
# What's new

Following a conversation with @DrZlo13 and @Skorpionm in #146 :

This adds integration of MemBrowse into the CI.
Now for every change in a PR and push to dev a memory report will be generated and sent to membrowse.com
To configure the setup:
1. Go to [membrowse.com](https://membrowse.com) and create an account and a project.
2. Copy the API key
3. Create a GitHub secret named MEMBROWSE_API_KEY and paste there the API key value.
4. In the project settings in the MemBrowse portal, set the project as publicly available so the dashboard will be available to all
5. Optional: run the "Onboard to MemBrowse" workflow manually to upload the last N commits builds

# Verification 

- Create a PR with a change and see that it works

# Contributor checklist


.github/workflows/build.yml
```diff
@@ -56,6 +60,7 @@ jobs:
       with:
         path: firmware
         submodules: recursive
+        fetch-depth: 2
```
electricalgorithm, line 63: Out of curiosity: Why the last two, but not only HEAD?

.github/workflows/build.yml
```diff
@@ -108,7 +113,30 @@ jobs:
...
+        cp ../../pico-sdk/src/rp2_common/pico_crt0/rp2350/memmap_default.ld ldscripts/
+        cp "$(find . -name pico_flash_region.ld -print -quit)" ldscripts/
+
+    - name: Run MemBrowse analysis
+      uses: membrowse/membrowse-action@v1
+      env:
+        GIT_DIR: ${{github.workspace}}/firmware/.git
+        GIT_WORK_TREE: ${{github.workspace}}/firmware
+      with:
+        target_name: rp2350-flipperone
```
electricalgorithm, line 133: Is it defined somewhere in your backend, or do we define the name here? I mean, what happens if we'd change it to something stupid like `alohaboard`.

README.md
```diff
@@ -1,5 +1,7 @@
 # Flipper One MCU Firmware
 
+[![MemBrowse](https://membrowse.com/badge.svg)](https://membrowse.com/public/flipperdevices/flipperone-mcu-firmware)
```
electricalgorithm, line 3: I guess maintainers have to change the link to correct MemBrowse project as well?

.github/workflows/build.yml
michael-membrowse: MemBrowse uses the commit hash of the parent commit to chain the reports. So the parent commit metadata is needed.

.github/workflows/build.yml
michael-membrowse: You name it here. The target is then automatically created in the backend on the first run. You can name it `alohaboard` if you want :-)

README.md
michael-membrowse: This is the correct path. The maintainers need to create an account on the MemBrowse portal and create a project with this repo (https://github.com/flipperdevices/flipperone-mcu-firmware). The link is basically https://membrowse.com/public + github repo path.

Currently there is no project created, so that's why the link seems broken.

.github/workflows/build.yml
electricalgorithm: Ty!

electricalgorithm:
LGTM. Thanks for explaining the set-up as well. It looks like the project is free for open source, and since PR introduces the most important steps; do we have an ETA for enabling membrowse @DrZlo13 @Skorpionm? It would be really cool to have it in PR level comments, so that one can introduce hard requirements for CLI tools or so on.

DrZlo13 (approved):

DrZlo13:
Sorry for the delay, our devops team was busy.

PR #149, opened 2026-05-31
configure the nvm reserve section in the linker script

michael-membrowse:
# What's new

Reserve the flash region used by the NVM/KV store in the linker configuration.

`furi_hal_nvm.c` carves `FURI_HAL_NVM_BANK_DEFAULT_SIZE` (128 KiB) out of flash
at runtime, but the linker was unaware of it. This adds a `NOLOAD` reservation
(`targets/f100/flash_nvm_reserve.ld`), which guarantees that:

1. The build fails if firmware grows large enough to collide with the flash
   reserved for the NVM, rather than overlapping it at runtime.
2. The reserved region appears in the ELF, so MemBrowse accounts for it and no
   longer reports those 128 KiB as free flash.

# Verification 

- Local build and elf inspection

# Contributor checklist


DrZlo13 (approved):

DrZlo13:
Thanks, this is a solid catch. LGTM

PR #165, opened 2026-06-03
Headphones srv: add debounce check connect headphone

### LINKED ISSUE #164: Headphones: debounce on insert

Skorpionm:
# What's new

- Headphones srv: add debounce check connect headphone

# Verification 

# Contributor checklist


applications/services/headphones/headphones.c
```diff
@@ -5,7 +5,8 @@
 
 #define TAG "HeadphonesSrv"
 
-#define HEADPHONES_TIMEOUT_UPDATE_MS (100)
+#define HEADPHONES_TIMEOUT_UPDATE_MS             (100)
+#define HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER 10 // Timeout = HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER * HEADPHONES_TIMEOUT_UPDATE_MS
```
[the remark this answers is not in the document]
Skorpionm: fix

applications/services/headphones/headphones.c
```diff
@@ -19,6 +20,7 @@ struct Headphones {
     FuriEventLoop* event_loop;
     FuriEventLoopTimer* timer;
     FuriPubSub* event_pubsub;
+    uint32_t debounce_counter;
 };
```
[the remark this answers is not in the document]
Skorpionm: fix

applications/services/headphones/headphones.c
```diff
@@ -5,7 +5,8 @@
 
 #define TAG "HeadphonesSrv"
 
-#define HEADPHONES_TIMEOUT_UPDATE_MS (100)
+#define HEADPHONES_TIMEOUT_UPDATE_MS             (100U)
+#define HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER (10U) // Timeout = HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER * HEADPHONES_TIMEOUT_UPDATE_MS
```
electricalgorithm: typo: s/COUTER/COUNTER

applications/services/headphones/headphones.c
```diff
@@ -49,34 +51,47 @@ static void headphones_custom_event_callback(uint32_t events, void* context) {
...
     }
 
     if(events & HeadphonesEventTypeIsrDisconnected) {
-        if(headphones_update(&hp_status)) {
-            HEADPHONES_SRV_DEBUG(TAG, "Headphones status changed: %08b", hp_status);
-            HeadphonesEvent event = {
-                .hp_status = hp_status,
-            };
-            furi_pubsub_publish(instance->event_pubsub, &event);
+        if(instance->debounce_counter >= HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER) {
```
electricalgorithm: Shouldn't we have a different debounce counter (perhaps something less than 10) since disconnection will cause problems for the tasks not reading the IO? Imagine someone plugging the jack and taking it back in one second. The system will still think that it's plugged in.

applications/services/headphones/headphones.c
```diff
@@ -19,6 +20,7 @@ struct Headphones {
     FuriEventLoop* event_loop;
     FuriEventLoopTimer* timer;
     FuriPubSub* event_pubsub;
+    uint32_t debounce_counter;
```
electricalgorithm, line 23: It looks like the headphones driver already have some kind of debouncing logic in place: https://github.com/flipperdevices/flipperone-mcu-firmware/blob/dev/lib/drivers/headphones/headphones.c#L157-L170

In total, a change will only apply after HEADPHONES_CHECK_CONNECT_DEBOUNCE_COUTER * HEADPHONES_TIMEOUT_UPDATE_MS + HEADPHONES_ADC_DEBOUNCE_MS * HEADPHONES_ADC_DOUBLE_CHECK = 1.2 seconds in the worst case. Are we OK with this latency?

applications/services/headphones/headphones.c
```diff
@@ -5,7 +5,8 @@
 
 #define TAG "HeadphonesSrv"
 
-#define HEADPHONES_TIMEOUT_UPDATE_MS (100)
+#define HEADPHONES_TIMEOUT_UPDATE_MS             (100U)
```
electricalgorithm: It is a tick interval, right? I believe it would be clear if we named it INTERVAL or something.

applications/services/headphones/headphones.c
electricalgorithm, line 48: The logic for connecting and disconnecting is different for the callback. We are not expecting N stable ticks for the disconnection loop. I believe it's important to comment on it in the codebase. it's not clear for the eye.

electricalgorithm (commented):
I searched a bit online and found out that why you have used counter-based debouncing instead of timer-based or bitshift-based ones. I learnt it from your code, and thanks for that.

applications/services/headphones/headphones.c
DrZlo13: Let me clarify why this logic is needed in the first place.

Currently, the codec can send an audio signal to the headphones even when they're unconnected, at a fairly large amplitude (3.3V p2p). This triggers our headphone detection system, resulting in false positives.

Latency isn't particularly critical here — we've increased it for now and will fine-tune it based on the test results.

applications/services/headphones/headphones.c
Skorpionm: No, it won't. The events come from interrupts that are triggered when the connector is inserted and removed.

applications/services/headphones/headphones.c
Skorpionm: This debouncer reads keystrokes, reads ADC data, and works independently of the event timer.

applications/services/headphones/headphones.c
Skorpionm: Maybe

applications/services/headphones/headphones.c
Skorpionm: Yes, it is different, we need to determine exactly that the plug is inserted, and disconnection is always determined correctly

applications/services/headphones/headphones.c
Skorpionm: Yes, there is a mistake.

Skorpionm: fix

applications/services/headphones/headphones.c
Skorpionm: fix

DrZlo13 (approved):

Issue #170, opened 2026-06-04
[Bug Report] New CLI Breaks the Shell Functionality for Pico 2


electricalgorithm:
The new CLI (PR #127) introduces separate-thread execution for commands that don't have the CliCommandFlagUseShellThread flag. Commands like top, log, power, power_consumption run in a dedicated FuriThread, and the shell thread blocks in furi_thread_join() waiting for the command thread to terminate.

The command's loop exits correctly on Ctrl+C (detected by cli_is_pipe_broken_or_is_etx_next_char()), but the CLI freezes immediately after — the prompt is never restored.

## Reproduction
1. Connect via USB CDC (screen / minicom / picocom)
2. Run top, log, power, or power_consumption
3. Press Ctrl+C to exit the loop
4. Observed: command exits, but the CLI prompt is never restored — the shell is permanently frozen

## Workaround
The workaround is found by @HobbitTheCat in #140.
> Add CliCommandFlagUseShellThread to the command registration flags — this runs the command inline in the shell thread (like the old CLI), bypassing the separate-thread path entirely.

## Root Cause Analysis with AI

Priority inversion in the thread scrub mechanism. The shutdown sequence is:

1. Command thread callback returns → furi_thread_body() suspends itself via vTaskSuspend(NULL) after putting itself in the scrub queue
2. Scrub thread (priority FuriThreadPriorityInit = 4) should pick up the message, call vTaskDelete(), and set state to Stopped
3. Shell thread (priority FuriThreadPriorityNormal = 16) polls with furi_delay_tick(2) waiting for Stopped
The scrub thread at priority 4 never gets CPU because the shell thread at priority 16 stays ready (it wakes from vTaskDelay(2) every 2ms). Any other service thread at priority 16 that stays runnable also starves the scrubber. The shell thread loops forever in furi_thread_join().

The scrub thread at priority 4 never gets CPU because the shell thread at priority 16 stays ready (it wakes from vTaskDelay(2) every 2ms). Any other service thread at priority 16 that stays runnable also starves the scrubber. The shell thread loops forever in furi_thread_join().

electricalgorithm:
## Root Cause Analysis

Find the AI's summary on the root cause analysis:

[...]

PR #171, opened 2026-06-04
fix(cli): resolve CLI hang on RP2350 due to priority inversion

electricalgorithm:
# What's new

This changeset fixes CLI hang on RP2350 (dual-core) where `furi_thread_join()` loops forever after a command thread exits. Root cause was a priority inversion — the Init task that runs thread cleanup couldn't preempt GuiSrv on dual-core SMP. The problem already is tracked on issue #170.

To resolve the explained problem, I had to change `FuriThreadPriorityInit` (4) to `FuriThreadPriorityNormal` (16) for the Init task, so that it runs at the same priority as service threads. I assume it's safe because Init only runs briefly during boot, then blocks forever on the (normally empty) scrub queue.

Something I'm not sure: This repo only has one target (`targets/f100/`), so changing its `main.c` affects all builds — including those running on plain Pico 2 dev boards for testing. If the project ever adds separate targets, the Init task priority should probably live in a shared init path rather than a board-specific target.

# Verification

- Before the fix, run any CLI tool and send CTRL+C with USB CDC. Validated that the issue exists and the CPU hangs.
- Apply the fix, build and flash to the RPi Pico 2. Run any CLI tool and send CTRL+C with USB CDC. Validated that the shell comes back and we're online with the CPU.

# Contributor checklist

### DIFF
--- a/targets/f100/src/main.c
+++ b/targets/f100/src/main.c
@@ -47,7 +47,7 @@ int main(void) {
     furi_hal_init_early();
 
     FuriThread* main_thread = furi_thread_alloc_ex("Init", 4096, init_task, NULL);
-    furi_thread_set_priority(main_thread, FuriThreadPriorityInit);
+    furi_thread_set_priority(main_thread, FuriThreadPriorityNormal);
     furi_thread_start(main_thread);
 
     // somehow openocd fucks up the multicore reset


HobbitTheCat:
Can confirm, the command now exits correctly with this fix

Skorpionm:
Sorry, but I don't understand what you're talking about. Are you saying that ctrl+c doesn't terminate top and it's impossible to restart anything??? I haven't seen this behavior, and I wouldn't publish code that reproduces it... Tell me about your setup.

Skorpionm:
and the second core on the 2350 is not used at all now, it is left for user scenarios, and simply by using only 1 core the chip consumes less electricity

electricalgorithm:
> Are you saying that ctrl+c doesn't terminate top and it's impossible to restart anything???

Yes, exactly. I am sure that it works on your demo device, but for us, it hangs all the time.

electricalgorithm:
> Tell me about your setup.


- Board: Pico 2 RP2350
- Host: Mac OS
- Connection: USB (not directly from UART0/1)
- Build: GCC Arm toolchain + CMake

Here's the steps:

- Latest dev, build it with CMake (No VSCode) and flash it.
- Use USB to connect Mac OS and the board (not from UARTO/1s)
- Start picocom/minicom session
- Run any tool, wait a bit and send ctrl+c -- the MCU hangs, no shell anymore

electricalgorithm:
@HobbitTheCat Can you also provide your setup to have more data?

HobbitTheCat:
Here is my setup and steps to reproduce:
- Board: Raspberry Pi Pico 2 W (RP2350)
- Debugger/Flash tool: Raspberry Pi Pico (RP2040) running picoprobe
- Host: NixOS
- Connection: USB CDC
- Build Environment: GCC ARM toolchain + CMake, isolated inside nix-shell

Steps to reproduce:
- Built the latest dev branch inside the nix-shell environment and flashed it via picoprobe
- Connected the board to the host via USB
- Opened the serial session using `screen`/`minicom`/`picocom`
- Ran a CLI command, waited a moment, and sent Ctrl+C

MCU freezes instantly and the shell becomes completely unresponsive

Skorpionm:
Yes, I reproduced the same problem on the Pico 2 board and the Pimoroni board.
It’s really interesting why this doesn't happen on our boards... but yes, the fact is the thread just hangs there, waiting to be shut down.

I also created a small patch to ensure the firmware loads correctly on the Pico 2 and to enable debugging and log output.
https://github.com/flipperdevices/flipperone-mcu-firmware/pull/172

Skorpionm:
I’ve found the problem, and it lies elsewhere entirely; your fix won’t be accepted because it alters the system's operating logic.

The problem is the lack of a physical display; our frame synchronization relies on a physical interrupt from the display's TE pin, which signals that a new frame can be loaded. Since you don't have a display, there's nothing to toggle that pin—I'll think about how to fix this!
image

HobbitTheCat:
Thanks for tracking this down! This behaviour affects multiple subsystems as well, for example, the missing display also blocks correct handling of `GUI_EVENT_FLAG_REDRAW`. That was preventing me from getting any screen output to the console. Commenting out `display_jd9853_qspi_write_buffer(gui->display, canvas_get_data(gui->render_canvas), width * height)` in `gui.c` makes things work correctly on my end, so I'll wait for the proper fix on that side.

Skorpionm:
check https://github.com/flipperdevices/flipperone-mcu-firmware/pull/173
image
rtt log 
image

If everything is fine, please close the PR and the associated issues.

DrZlo13:
Hello. I believe this issue is resolved by #173. Сan you please confirm and close this PR if so?

electricalgorithm:
I can confirm that the issue is resolved in dev branch. Thanks!

Issue #140, opened 2026-05-26
[Bug Report] `furi_hal_serial_async_rx_stop` causes MCU freeze on RP2350


electricalgorithm:
Calling `furi_hal_serial_async_rx_stop()` after `furi_hal_serial_async_rx_start()` freezes the entire MCU when the RX callback properly drains the FIFO. No crash output is produced — CLI (USB CDC) becomes unresponsive, no further output appears. The freeze occurs inside or immediately after `async_rx_stop` — any `printf` placed after the call never executes.

### Steps to Reproduce

Apply the attached patch and run the `uart_repro` CLI command on the Raspberry Pi Pico 2.

[uart_repro_bug.patch](https://github.com/user-attachments/files/28272312/uart_repro_bug.patch) - This patch easily applies to the current dev branch (ad122609).

Find what happens and when marked next to the reproducer's calls:
```c
furi_hal_serial_async_rx_start(h, false);     // OK — prints "async_rx_start OK"

furi_delay_ms(100);

furi_hal_serial_async_rx_stop(h);              // Freeze — next line never prints
printf("async_rx_stop OK");                    // Never reached
```

Console:
```txt
Firmware version: dev r94 (cae4bed-dirty built on 2026-05-26)

>: ?
Commands available:
?                             led
clock_out                     log
expander_ext                  otp
free                          power
free_blocks                   power_consumption
gpio                          top
help                          uart_repro
i2c                           uptime
>: uart_rep
uart_repro

>: uart_repro
UART repro: starting test
UART repro: acquire OK
UART repro: init OK
UART repro: callback set OK
UART repro: async_rx_start OK

```
There it freezes. I have to reboot the hardware.

Here's the details about my target, Pico SDK version and my UART set-up.
- **Target:** Raspberry Pi Pico 2
- **SDK:** Pico SDK v2.2.0
- **UART:** UART1 at 230400 baud, RX=GPIO9 (patched targets/ to work), TX=GPIO36 (not patched, not used). On the other end of the uart, there were an UART-to-USB adaptor but I don't think it relates.

### What I Tried

| Attempt | Result |
|---|---|
| Full stop: `async_rx_stop` + `deinit` + `control_release` | Freeze |
| Remove `control_release`, keep `async_rx_stop` + `deinit` | Freeze |
| Remove `deinit`, keep only `async_rx_stop` | Freeze |

---

[...]

PR #175, opened 2026-06-07
[CLI, MCU <-> CPU] feat(bootlog): add bootlog service for Linux boot log

electricalgorithm:
# What's new

This PR adds the bootlog service and its CLI to capture the Linux CPU's boot log from UART0 (can be changed; idk which UART is used, but since UART1 is used, I go with UART0) and stores it in a ring buffer. The CLI exposes the following commands: 
- bootlog start
- bootlog stop (does not work yet, see #140)
- bootlog show
- bootlog clear
- bootlog status

An observer/subscriber API for GUI access is also provided. The logic works as follows:
1. UART0 ISR gets the characters and adds them to the FuriStreamBuffer,
2. When the ISR is finished, FuriEventLoop reads the buffer and saves it to the app's ring buffer,
3. With the observer pattern, each subscriber is notified (CLI is one of them).

For the GUI (or others in future), services can subscribe to `BootlogNotification` events via `bootlog_subscribe()` and receive a message with `new_head` and `bytes_available` whenever new data arrives. I evaluated a "cursor" based API for full-read (not only event-based, but read the full ring whenever it is needed), but did not add the implementation in this PR. If we think that it is needed, we can do it later on. One can find the implementation on the following patch:

[bootlog-cursor-api.patch](https://github.com/user-attachments/files/28688710/bootlog-cursor-api.patch)

The missing cursor-based API might help us to get the full bootlog after the subscriptions, when the user wants to read the full log even without the GUI being subscribed. I don't know if this is a valid scenario, tho.

**Important**!
`furi_hal_serial_async_rx_stop()` freezes the MCU on RP2350 (see #140). Calling `bootlog stop` triggers this freeze and kills USB CDC. Workaround: start the service once and leave it running for the device's lifetime. Do not call stop until the underlying HAL bug is fixed.

One can see the issue on the attached screenshot video.

# Verification

https://github.com/user-attachments/assets/2582c991-2b19-4065-9875-73d0c3f4052e

1. Connect host UART TX to RP2350 UART0 RX (GPIO 33 on Flipper One, GPIO 1 on Pico 2)
2. Flash firmware, connect via USB CDC
3. Run bootlog start: confirms UART acquisition and baud rate
4. Send data from host at 230400 baud
5. Run bootlog show: displays captured data

# Contributor checklist

### DIFF
--- a/applications/applications.c
+++ b/applications/applications.c
@@ -6,6 +6,8 @@ const char* FLIPPER_AUTORUN_APP_NAME = "";
 extern int32_t haptic_srv(void* p);
 extern int32_t test_peref_srv(void* p);
 extern int32_t input_srv(void* p);
+extern int32_t uart_echo_app(void* p);
+extern int32_t bootlog_srv(void* p);
 extern int32_t input_touch_srv(void* p);
 extern int32_t i2c_intercom_srv(void* p);
 extern int32_t i2c_negotiator_srv(void* p);
@@ -34,6 +36,7 @@ extern void power_cli(PipeSide* pipe, FuriString* args, void* context);
 extern void power_consumption_cli(PipeSide* pipe, FuriString* args, void* context);
 extern void led_cli(PipeSide* pipe, FuriString* args, void* context);
 extern void uart_echo_cli(PipeSide* pipe, FuriString* args, void* context);
+extern void bootlog_cli(PipeSide* pipe, FuriString* args, void* context);
 
 const FlipperInternalApplication FLIPPER_SERVICES[] = {
     {
@@ -71,6 +74,13 @@ const FlipperInternalApplication FLIPPER_SERVICES[] = {
     //     .stack_size = 2048,
     //     .flags = FlipperInternalApplicationFlagDefault,
     // },
+    {
+        .app = bootlog_srv,
+        .name = "BootlogSrv",
+        .appid = "bootlog",
+        .stack_size = 2048,
+        .flags = FlipperInternalApplicationFlagDefault,
+    },
     {
         .app = input_touch_srv,
         .name = "InputTouchSrv",
@@ -254,6 +264,12 @@ const FlipperInternalCommandApplication FLIPPER_CLI_COMMANDS[] = {
         .callback = uart_echo_cli,
         .name = "uart_echo",
         .stack_size = 1024 * 2,
+        .flags = CliCommandFlagParallelSafe,
+    },
+    {
+        .callback = bootlog_cli,
+        .name = "bootlog",
+        .stack_size = 1024,
         .flags = CliCommandFlagParallelSafe,
     },
 };
--- a/applications/services/bootlog/bootlog.c
+++ b/applications/services/bootlog/bootlog.c
@@ -1,0 +1,447 @@
+#include "bootlog.h"
+
+#include <furi.h>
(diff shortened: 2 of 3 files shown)


electricalgorithm:
Questions to maintainers:
- Is UART0 the one CPU connected to?
- Do we want to see full buffer on show CLI command or just the new messages coming?
- Should this service be online (started) for the full lifetime or do you see value of start/stop logic?

Skorpionm:
The Linux boot console is now available on Debug Connector One. The Linux console operates at 1,500,000 baud (1.5 Mbaud). It supports both data transfer and reception.

Skorpionm:
What I'm getting at is that the data from the Linux terminal is coming in very quickly and from a lot of data. Your implementation (judging by the video) only returns the last line of the log, which isn't informative. Plus, if you look closely, if the new line was shorter than the previous one, the new line and the end of the old one are displayed. Regarding the function to stop using UART, I'll check it out; it worked.

Skorpionm:
event_loop has implementations for queue processing, subscriptions, and notifications; there's no need to invent new implementations. And your service, even when not in use, simply constantly uses 12 kilobytes of RAM, so why such an implementation?

Skorpionm:
https://github.com/flipperdevices/flipperone-mcu-firmware/pull/176
check plz, fix "furi_hal_serial_async_rx_stop"

electricalgorithm:
> The Linux boot console is now available on Debug Connector One. The Linux console operates at 1,500,000 baud (1.5 Mbaud). It supports both data transfer and reception.

I checked it, thanks for sharing. Plus, I updated the baud rate to be 1.5MBps.

> What I'm getting at is that the data from the Linux terminal is coming in very quickly and from a lot of data. Your implementation (judging by the video) only returns the last line of the log, which isn't informative.

You're right. I did not plan the service / CLI to support live tailing, but now the new version does with a new subcommand `bootlog stream`. GUI and any other client must query the buffer when there's a new notification and save it / use it then. This is exactly what bootlog stream does now. It's better to have it this way (IMO) since we do not need more than a buffer size.

> Plus, if you look closely, if the new line was shorter than the previous one, the new line and the end of the old one are displayed.

Yes, you're right. There wasn't any CR/LF handling logic there. I added it in the new revision.

> event_loop has implementations for queue processing, subscriptions, and notifications; there's no need to invent new implementations.

TBH, there's no documentation regarding the FURI library. I checked it, and now implemented using the PubSub there. I hope this is what you mentioned. 

> And your service, even when not in use, simply constantly uses 12 kilobytes of RAM, so why such an implementation?

Right. That's my bad, it was the middle of the night. I changed the implementation to allocate the memory on start and free it on stop subcommands. That'd be better.

> https://github.com/flipperdevices/flipperone-mcu-firmware/pull/176
check plz, fix "furi_hal_serial_async_rx_stop"

Done. It is working. Thanks!

electricalgorithm:
Please find my automated test case with ~28KB bootlog simulation data.

[bootlog_simulator.py](https://github.com/user-attachments/files/28934681/bootlog_simulator.py)

Find the demo usage of the bootlog service and CLI tools:

https://github.com/user-attachments/assets/134d4d67-cd9b-46d7-838d-32abcf6638f7

electricalgorithm:
What's the current status? No reviews yet.

PR #179, opened 2026-06-08
CLI: display emulation

### LINKED ISSUE #142: Feature Request: Headless debugging with virtual display emulation via USB CLI
Hello Flipper Devices team and contributors,

In issue #135, the maintainers noted that building a hardware test bench is currently impractical since the control board schematic hasn't been finalized yet. To enable hardware-less core development on generic RP2350 dev boards in the meantime, I implemented a Virtual Display Emulation interface over USB CLI.

#### How it works:
1.  I created an alternative implementation of the display interface (`display_jd9853_usb_virtual.c/.h`) mimicking the `original display_jd9853_qspi.c` driver. Instead of using HSTX and hardware DMA transfers, it captures the raw framebuffer via memcpy inside write_buffer and releases the synchronization semaphores, keeping the core FreeRTOS tasks unblocked.
2. The abstraction layers are fully preserved. The system registers and interacts with the virtual display as a standard system record, unaware of the underlying injection.
3. To prevent USB CDC buffer overflows and TX task starvation, I added a standalone screen command inside `services/cli`. It hooks up to the virtual driver's frame buffer.
4. To fit the screen into a standard console window, the frames are rendered 1:1 using Unicode Braille patterns.
5. The entire feature is wrapped under a conditional compilation flag (`ENABLE_USB_DISPLAY_EMULATION`) inside `CMakeLists.txt`. When turned OFF, the debug code is completely stripped, causing zero overhead or regressions in the product build.

#### Current State & Verification 
I have fully verified the build. Memory management is clean, and there are no race conditions or deadlocks when exiting the viewer.

Image

If the core team is interested in integrating this feature for contributors who want to work on firmware without the screen hardware, I can open a Pull Request.

#### Next Steps

Once the display rendering is settled, the next logical step would be implementing virtual button injection via the CLI.

Looking forward to your feedback.

HobbitTheCat:
# What's new

- **Closes #142**
- **Dependency Inversion**:  (per @DrZlo13  feedback) Added `gui_add_framebuffer_callback` / `gui_remove_framebuffer_callback` to the GUI service (`gui.c`), following the dependency inversion pattern from Flipper Zero firmware, adapted for FlipperOne's setup.
- **`screen` CLI Command**: Implemented a screen streaming command inside `applications/services/cli/commands/` bound to the `ENABLE_USB_DISPLAY_EMULATION` conditional compilation flag (disabled by default).
- **In-Terminal Control**: Integrated an asynchronous keyboard intercept loop using `getchar()` and `CliAnsiParser`. It maps inputs into software `InputEvent` definitions, over a single VCP/UART CLI pipeline.

Display frames are rendered 1:1 using Unicode Braille patterns. A one-character-per-pixel grayscale approach was evaluated but rejected, on a device with fine UI text, it requires significant terminal font downscaling to fit the screen, which makes the interface unreadable in practice. Braille encoding preserves pixel accuracy within a standard terminal window without font size adjustments.

# Verification 

1. Build the firmware with the display emulation feature flag enabled via CMake:
    ```bash
   cmake -DENABLE_USB_DISPLAY_EMULATION=ON ..
   make
   ```
2. Connect to the FlipperOne via USB VCP or UART CLI terminal
3. Run the streaming tool: `screen`
4. Press `Arrow Keys` or `W/A/S/D` to navigate the system menus, `1/2/3/4/5` for bottom buttons, `p` for push-to-talk, 
`s` for SW, `Enter/o` to confirm actions, and `Backspace/b` to move back. 
5. Press CTRL+C to exit the screen viewer.

# Contributor checklist

cli_screen_test


Skorpionm:
Honestly, I haven't seen how it works yet, but maybe I could just add a registered callback when the app launches from the cli and unregister it when it's shut down? So that the app always runs smoothly. There's one problem, though: when Linux starts, the VCP is disabled and control of the USB goes to Linux.

HobbitTheCat:
Regarding the lifecycle: yes, that's exactly how it works! The callback is registered when `screen` launches and unregistered when the command exits. `GuiSrv` uses a non-blocking `TryAcquire mutex`, so the main app loop always runs smoothly regardless of terminal throughput.

On the VCP handover concern: fair point for the final board, but the feature stays useful in a few key scenarios:
1. Hardware-less development:  A convenient way for contributors without screen hardware to render the Clay GUI and verify visual tasks on a generic RP2350 board.
2. UART fallback: Since the CLI abstracts transports via PipeSide, this should still work over CliUart even when Linux claims the USB controller.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+            else if (key_code == 'd' || key_code == 'D') emulated_key = InputKeyRight;
+
+            else if (key_code == '1') emulated_key = InputKey1;
+            else if (key_code == '2') emulated_key = InputKey2;
+            else if (key_code == '3') emulated_key = InputKey3;
+            else if (key_code == '4') emulated_key = InputKey4;
+            else if (key_code == '5') emulated_key = InputKey5;
+
+            else if (key_code == '\t' || key_code == 'x' || key_code == 'X') emulated_key = InputKeySw;
+            else if (key_code == 'p' || key_code == 'P') emulated_key = InputKeyPtt;
```
DrZlo13: I think switch-case would be much nicer here.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        line_buffer[line_pos++] = '|';
+        line_buffer[line_pos++] = '\r';
+        line_buffer[line_pos++] = '\n';
+        line_buffer[line_pos]   = '\0';
+
+        pipe_send(pipe, (uint8_t*)line_buffer, line_pos);
+    }
+}
```
DrZlo13: I don't like that there's no bounds check here — there should be at least one `furi_check` at the very end. Memory corruption is the most painful thing to debug in the embedded systems.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+
+    Gui* gui = furi_record_open(RECORD_GUI);
+    FuriPubSub* input_pubsub = furi_record_open(RECORD_INPUT_EVENTS);
+    
+    if (!gui || !input_pubsub) {
+        printf("Error: Required subsystems not found\r\n");
+        if (gui) furi_record_close(RECORD_GUI);
+        if (input_pubsub) furi_record_close(RECORD_INPUT_EVENTS);
+        return;
+    }
```
DrZlo13: `furi_record_open` guarantees that resource will be available, quote: "Suspends caller thread till record is available"

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+    char* line_buffer = malloc(line_buffer_size);
+    
+    if (!framebuffer || !line_buffer) {
+        printf("Error: Memory allocation");
+        if (framebuffer) free(framebuffer);
+        if (line_buffer) free(line_buffer);
+        furi_record_close(RECORD_GUI);
+        furi_record_close(RECORD_INPUT_EVENTS);
+        return;
+    }
```
DrZlo13: Our `malloc` also guarantees to return a valid memory.

CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ set(FIRMWARE_TARGET 100 CACHE STRING "Firmware target number")
 set(FIRMWARE_ORIGIN "Official" CACHE STRING "Firmware origin")
 set(TARGET_HARDWARE "f${FIRMWARE_TARGET}")
 
+option(ENABLE_USB_DISPLAY_EMULATION "Build with USB display emulation command in CLI" OFF)
```
DrZlo13: I think that there is no point in disabling such a useful feature.

DrZlo13 (changes requested):

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+            InputKey emulated_key = InputKeyMask;
+            CliKey key_code = ansi_res.result.key;
+            
+            if (key_code == CliKeyUp)          emulated_key = InputKeyUp;
+            else if (key_code == CliKeyDown)   emulated_key = InputKeyDown;
+            else if (key_code == CliKeyLeft)   emulated_key = InputKeyLeft;
+            else if (key_code == CliKeyRight)  emulated_key = InputKeyRight;
+            else if (key_code == CliKeyEsc)    emulated_key = InputKeyBack;
+
+            else if (key_code == '\r' || key_code == '\n' || key_code == ' ' || key_code == 'o') emulated_key = InputKeyOk;
```
DrZlo13: When I type "screen" and press Enter, "OK" is sent to the device after the first frame. 
I'm not sure, but it seems like there is a `\r` or `\n` left in the pipe after the command input, and the pipe should be flushed before use.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+    uint8_t* local_framebuffer;
+    FuriMutex* mutex;
+    bool has_new_frame;
+} CliScreenContext;
+
+static void cli_screen_framebuffer_callback(const uint8_t* data, size_t width, size_t height, void* context) {
+    CliScreenContext* ctx = context;
+    if (!ctx || !data) return;
+
+    if (furi_mutex_acquire(ctx->mutex, 0) == FuriStatusOk) {
```
DrZlo13: I think we can wait a bit, 10-20ms will be ok.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,202 @@
...
+            break;
+        }
+
+        if (furi_mutex_acquire(screen_ctx.mutex, FuriWaitForever) == FuriStatusOk) {
+            if (screen_ctx.has_new_frame) {
+                screen_ctx.has_new_frame = false;    
+                cli_command_screen_render_braille(pipe, screen_ctx.local_framebuffer, line_buffer, width, height);
+            }
+            furi_mutex_release(screen_ctx.mutex);
+        }
```
DrZlo13: Because of this approach (taking the mutex before reading and `furi_delay` after), we're dropping a lot of frames.

It would be better to protect only the frame read and write with the mutex, and replace `furi_delay` with waiting on an event flag, delivered from `cli_screen_framebuffer_callback`.

Ideally, we'd use an event loop attached to the pipe (`pipe_attach_to_event_loop`) and to event flag, which would let us handle everything asynchronously.

DrZlo13 (changes requested):

DrZlo13:
Sorry for all the comments! Great feature and solid code overall — it just needs a little bit of polishing.

HobbitTheCat:
Just pushed the first two batches of changes:
1. Replaced the `if-else` chain with a `switch` block, removed redundant validation, fixed the residual pipe characters, and added explicit `furi_check` memory bounds validation.
2. Cleaned up the conditional compilation flags across the related build files.

The frame-dropping fix is next. Will push it soon and post an update once it's ready for review.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,222 @@
...
+                cli_screen_emulate_click(input_pubsub, emulated_key);
+            }
+        }
+    }
+    return false;
+}
+
+static void cli_command_screen_render_braille(PipeSide* pipe, const uint8_t* fb, 
+    char* line_buffer, size_t line_buffer_size, size_t width, size_t height) {
+    printf("\033[H");
```
Skorpionm: some improvements to get away from unclear magical meanings
replace with printf(ANSI_CURSOR_POS("1", "1")); //move the cursor to position 1.1 without clearing the screen

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,222 @@
...
+    CliAnsiParser* ansi_parser = cli_ansi_parser_alloc();
+
+    printf("Controls: Arrows = Navigate, Enter = Ok, Backspace = Back\r\n");
+    printf("Press CTRL+C to stop... \r\n");
+    furi_delay_ms(500);
+
+    gui_add_framebuffer_callback(gui, cli_screen_framebuffer_callback, &screen_ctx);
+    gui_update(gui);
+
+    printf("\033[2J\033[H\033[?25l");
```
Skorpionm: printf(ANSI_ERASE_DISPLAY(ANSI_ERASE_ENTIRE)); // Clear display
printf(ANSI_CURSOR_POS("1", "1")); //move the cursor to position 1.1 without clearing the screen
\033[?25l Enable/disable cursor. I don't think this command exists yet. It might be useful if you add it.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,222 @@
...
+                screen_ctx.has_new_frame = false;    
+                cli_command_screen_render_braille(pipe, screen_ctx.local_framebuffer, line_buffer, line_buffer_size, width, height);
+            }
+            furi_mutex_release(screen_ctx.mutex);
+        }
+        
+        furi_delay_ms(40);
+    }
+
+    printf("\033[2J\033[H\033[?25h");
```
Skorpionm: the same

HobbitTheCat:
@DrZlo13 @Skorpionm , I pushed the final batch. Here's a summary of what changed:

Event-driven loop: Removed the furi_delay_ms(40) polling entirely. The command now blocks on furi_event_loop_run() and wakes only on actual input or a new frame signal.
Code cleanup: switch-case key mapping, furi_check bounds validation, ANSI macros via cli_ansi (also added CURSOR_HIDE/CURSOR_SHOW there for general use), removed the ENABLE_USB_DISPLAY_EMULATION flag.

Thanks again for the thorough review, ready for another look!

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,223 @@
...
+
+    char* linebuffer;
+    size_t linebuffer_size;
+
+    FuriMutex* mutex;
+} CommandScreen;
+
+static void cli_command_screen_gui_update_callback(const uint8_t* data, size_t width, size_t height, void* context) {
+    CommandScreen* instance = context;
+    if (!instance || !data) return;
```
DrZlo13: Throughout this and the following code, the `if (!something) return;` pattern is used.

If you set the context yourself, it can never be null, and if it suddenly is null — that's a clear bug, and it should be reported, not handled. Use `furi_check`.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,223 @@
...
+
+    instance->framebuffer = malloc(instance->framebuffer_size);
+    instance->linebuffer = malloc(instance->linebuffer_size);
+
+    instance->mutex = furi_mutex_alloc(FuriMutexTypeNormal);
+
+    pipe_attach_to_event_loop(instance->pipe, instance->event_loop);
+    pipe_set_callback_context(instance->pipe, instance);
+    pipe_set_data_arrived_callback(instance->pipe, cli_command_screen_click_callback, FuriEventLoopEventFlagEdge);
+    pipe_set_broken_callback(instance->pipe, cli_command_screen_pipe_broken_callback, FuriEventLoopEventFlagOnce);
```
DrZlo13, line 180: This part really worries me.

If you close the CDC connection while screen is running, the MCU goes into a Hard Fault. While debugging, I traced it down to exactly this spot.

The problem is that the `PipeSide* pipe` we receive already has its own arrived/broken handlers and is already connected to its own event_loop — but this code overwrites all of that.

I honestly don't know the right way to rewrite this, so I'll bring in the people who implemented `Pipe`. We also clearly need `furi_check` guards against overwriting the callbacks and the event in the pipe — that would have saved me a couple of hours. That part, though, we can implement on our side.

applications/services/cli/cli_command_screen.c
```diff
@@ -0,0 +1,223 @@
...
+            InputKey emulated_key = InputKeyMask;
+
+            switch ((int) ansi_res.result.key) {
+                case CliKeyUp:    case 'W': case 'w': emulated_key = InputKeyUp; break;
+                case CliKeyDown:  case 'S': case 's': emulated_key = InputKeyDown; break;
+                case CliKeyLeft:  case 'A': case 'a': emulated_key = InputKeyLeft; break;
+                case CliKeyRight: case 'D': case 'd': emulated_key = InputKeyRight; break;
+
+                case CliKeyEsc: case 0x7F: case 0x08: case 'B': case 'b': emulated_key = InputKeyBack; break;
+                case '\r': case '\n': case ' ': case 'O': case 'o': emulated_key = InputKeyOk; break;
```
DrZlo13, line 131: I think the `'\r'` here is unnecessary — some terminals send `\r\n` on Enter, which would trigger two "OK" presses in a row.

DrZlo13 (changes requested):

applications/services/cli/cli_command_screen.c
HobbitTheCat: Removing \r entirely would break terminals like minicom that send only \r on Enter. I can add simple deduplication to suppress the second event when \r\n arrives in sequence.

applications/services/cli/cli_command_screen.c
HobbitTheCat: Interesting, I didn't catch this during testing. I'll wait for the input from the `Pipe` authors before touching this part.

applications/services/cli/cli_command_screen.c
DrZlo13: Yes, I agree.

applications/services/cli/cli_command_screen.c
DrZlo13: @portasynthinca3 @gsurkov can you help us here?

DrZlo13: Unfortunately, that part of the team is really _busy_ right now. I think I'll merge this pull request as is and just open an issue for this problem. Thanks, this is good work.

DrZlo13 (approved):

NickVolynkin:
Hi @HobbitTheCat! Your contribution was featured in [Flipper One Dev Log #7](https://docs.flipper.net/one/dev-log/7#terminal-based-screen-viewer).

HobbitTheCat:
That's awesome, thank you for the mention! Happy to see it land in the dev log.

Issue #51, opened 2026-03-03
Debugging using the FreeRTOS plugin shows garbage in the call stack


DrZlo13:
<img width="493" height="492" alt="Image" src="https://github.com/user-attachments/assets/55e0bba5-b2cd-4aea-92a8-6882c69dc359" />

PaulFirs:
You are waiting for the `furi_thread_catch` function to be at the beginning of the stack, but there is no corresponding offset in FreeRTOSConfig.h (you did this in flipper-zero):
```c
extern __attribute__((__noreturn__)) void furi_thread_catch(void);
#define configTASK_RETURN_ADDRESS (furi_thread_catch + 2)
```

There is a solution to simplify this approach with the naked attribute.
In thread.c, change `__attribute__((__noreturn__)) void furi_thread_catch(void)` on
```c
/** Catch threads that are trying to exit the wrong way - crash implementation */
static __attribute__((__noreturn__)) void furi_thread_catch_impl(void) {
    // If you're here it means you're probably doing something wrong
    // with critical sections or with scheduler state
    furi_crash("You are doing it wrong");
    __builtin_unreachable();
}

__attribute__((naked, __noreturn__)) void furi_thread_catch(void) { //-V1082
    asm volatile(
".cfi_undefined lr\n\t" /* tell DWARF unwinder: no return address */
"b furi_thread_catch_impl\n\t");
}
```

and in FreeRTOSConfig.h add
```c
extern __attribute__((naked, __noreturn__)) void furi_thread_catch(void);
#define configTASK_RETURN_ADDRESS furi_thread_catch
```

DrZlo13:
@PaulFirs Thank you, good catch about `configTASK_RETURN_ADDRESS`.

But, unfortunately, this doesn't solve all the problems. For some threads, I still get garbage instead of a stack.
Considering that in the openocd logs, when querying the stack of these threads, I see `warning: Invalid state, unable to determine sp alias, assuming msp`, I'm betting there's a problem with stack parsing in this architecture (Cortex-M33).

DrZlo13:
To avoid being unfounded:

<img width="362" height="518" alt="Image" src="https://github.com/user-attachments/assets/f4493c9f-3473-4787-82d2-623a6298de2a" />

PaulFirs:
Try using `naked` with `.cfi_undefined lr`. This should help the DWARF unwinder to stop at the thread boundary.

[...]

PR #182, opened 2026-06-11
freertos: drop the '+ 2' from configTASK_RETURN_ADDRESS (after corelibs naked trampoline)

NetVar1337:
> [!NOTE]
> **Draft — depends on the corelibs change landing first.** This is the second half of the #51 fix and should not be merged until `furi_thread_catch` becomes a naked trampoline in corelibs (flipperdevices/fbtng-corelibs#33). Merging this alone would drop the prologue-skip while the symbol is still an ordinary function, which makes the unwind *worse*, not better. Once corelibs merges I'll add the submodule bump to the merged commit.

Companion to the corelibs change for flipperdevices/flipperone-mcu-firmware#51.

`configTASK_RETURN_ADDRESS` currently uses `(furi_thread_catch + 2)` to skip the function prologue so the debugger doesn't unwind into it. Once `furi_thread_catch` is a `naked` trampoline with `.cfi_undefined lr` (corelibs), the `+ 2` hack is unnecessary and slightly misleading — the FDE itself now terminates the unwind. This drops the offset and updates the `extern` declaration to match the naked attribute.

Verified: firmware builds clean before and after (arm-none-eabi-gcc 14.2.1, pico-sdk 2.2.0, `[264/264]`, uf2 produced). With both changes applied, `thread apply all bt` stops at the thread entry instead of printing an artificial caller frame.

No hardware on hand (compile + DWARF/disassembly verification only); on-silicon behaviour of the combined fix is per @DrZlo13's 2026-06-03 confirmation in #51 that the patched OpenOCD path "works great".

Ref: #51


kakavaanen:
Confirmed on real RP2350 hardware alongside fbtng-corelibs#33 and raspberrypi/openocd#151 — with all three applied, thread apply all bt unwinds all 20 FreeRTOS tasks cleanly (previously GuiSrv/PowerMenuSrv/HapticSrv were garbage). Full details + output on #51.

PR #187, opened 2026-06-17
Cli: add dmesg

### LINKED ISSUE #185: Logs: add a ring buffer to logs for storing history
- Add a ring buffer to the logs to store log history.
- The buffer must be static and configurable.
- It must integrate through the standard `furi_log_add_handler`.
- There must be a CLI command to display the history.
- `dmesg` command.

Skorpionm:
# What's new

- Cli: add dmesg
- Toolbox: add circular_buffer
- Circular_buffer: add unit_test

# Verification 

- 
image

# Contributor checklist

### DIFF
--- a/applications/main/dmesg/dmesg_app.h
+++ b/applications/main/dmesg/dmesg_app.h
@@ -12,6 +12,7 @@ typedef struct DmesgApp DmesgApp;
 
 bool dmesg_app_get_log_data(DmesgApp* instance, uint8_t* data, size_t* size);
 void dmesg_app_update_read_index(DmesgApp* instance);
+void dmesg_app_restore_read_index(DmesgApp* instance);
 
 #ifdef __cplusplus
 }
--- a/applications/main/dmesg/dmesg_app.h
+++ b/applications/main/dmesg/dmesg_app.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <furi_hal_resources.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define RECORD_DMESG_APP "dmesg_app"
+
+typedef struct DmesgApp DmesgApp;
+
+bool dmesg_app_get_log_data(DmesgApp* instance, uint8_t* data, size_t* size);
+void dmesg_app_update_read_index(DmesgApp* instance);
+
+#ifdef __cplusplus
+}
+#endif
--- a/applications/main/dmesg/dmesg_app.h
+++ b/applications/main/dmesg/dmesg_app.h
@@ -1,6 +1,7 @@
 #pragma once
-
-#include <furi_hal_resources.h>
+#include <stdint.h>
+#include <stddef.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
--- a/applications/main/dmesg/dmesg_app.h
+++ b/applications/main/dmesg/dmesg_app.h
@@ -13,6 +13,8 @@ typedef struct DmesgApp DmesgApp;
 bool dmesg_app_get_log_data(DmesgApp* instance, uint8_t* data, size_t* size);
 void dmesg_app_update_read_index(DmesgApp* instance);
 void dmesg_app_restore_read_index(DmesgApp* instance);
+void dmesg_app_read_acquire(DmesgApp* instance);
+void dmesg_app_read_release(DmesgApp* instance);
 
 #ifdef __cplusplus
 }
--- a/applications/main/dmesg/dmesg_cli.c
+++ b/applications/main/dmesg/dmesg_cli.c
@@ -15,9 +15,6 @@ static void dmesg_cli_command_dmesg_help(void) {
 }
 
 static void dmesg_cli_show_log(PipeSide* pipe, FuriString* args) {
-    UNUSED(pipe);
-    UNUSED(args);
-
     bool wait_for_new_logs = false;
 
     if(furi_string_size(args) > 0) {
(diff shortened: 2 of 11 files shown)


applications/main/dmesg/dmesg_app.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <furi_hal_resources.h>
```
DrZlo13: `#include <stdint.h, stddef.h, stdbool.h>` will suffice.

applications/main/dmesg/dmesg_cli.c
```diff
@@ -0,0 +1,67 @@
...
+static void dmesg_cli_command_dmesg_help(void) {
+    printf("Usage:\r\n");
+
+    printf(" dmesg [w] [log level]\t - show the last log line, if \"w\" is provided, wait for new log lines\r\n");
+    printf(" \t\t\tLog level can be one of: default, none, error, warn, info, debug, trace\r\n");
+}
+
+static void dmesg_cli_show_log(PipeSide* pipe, FuriString* args) {
+    UNUSED(pipe);
+    UNUSED(args);
```
DrZlo13: really unused?

lib/toolbox/circular_buffer.c
```diff
@@ -0,0 +1,110 @@
...
+    /* Write data at head position */
+    const size_t head = cb->head;
+    size_t first = cb->size - head;
+    if(size <= first) {
+        memcpy(cb->data + head, data, size);
+        cb->head = (head + size) % cb->size;
+    } else {
+        memcpy(cb->data + head, data, first);
+        memcpy(cb->data, data + first, size - first);
+        cb->head = size - first;
```
DrZlo13, line 88: In the wrap-around branch, `memcpy(cb->data, data + first, size - first)` overflows `cb->data` when `size - first > cb->size` (i.e. a single write larger than the buffer). 

Not reachable from dmesg today (log fragments are capped at 256B), but circular_buffer is a reusable toolbox component with no guard on the contract.

lib/toolbox/circular_buffer.c
```diff
@@ -0,0 +1,110 @@
...
+
+void circular_buffer_release(CircularBuffer* cb) {
+    furi_mutex_release(cb->mutex);
+}
+
+size_t circular_buffer_bytes_available(const CircularBuffer* cb) {
+    const size_t head = cb->head;
+    const size_t tail = cb->tail;
+    return (head >= tail) ? (head - tail) : (cb->size - (tail - head));
+}
```
DrZlo13, line 45: "Empty" and "full" both map to `head == tail`, so real capacity is `size - 1`, but `circular_buffer_spaces_available` returns `size - bytes_available` (one too many). In `circular_buffer_write`, writing exactly `spaces_available` bytes fails the `size > space check`, takes the normal path, and lets head catch up to tail — so `bytes_available` reads 0 and the entire buffer silently empties.

applications/main/dmesg/dmesg_cli.c
```diff
@@ -0,0 +1,67 @@
...
+    dmesg_app_update_read_index(instance);
+
+    while(dmesg_app_get_log_data(instance, buffer, &size)) {
+        if(size != 0) {
+            printf("%.*s", (int)size, buffer);
+        }
+        size = DMESG_BUFFER_SIZE;
+    }
+    dmesg_app_restore_read_index(instance);
+    dmesg_app_read_release(instance);
```
DrZlo13, line 51: The dump holds `cb->mutex` across all printf calls. If the pipe/USB/CLI layer ever emits a `FURI_LOG_*`, the log handler re-enters `circular_buffer_acquire` on the same non-recursive mutex → same-thread self-deadlock, and since `furi_log.mutex` is already held, all system logging freezes (AB-BA).

Quick fix: copy the snapshot into a temp buffer under the lock, release the mutex, then printf from the copy — decouples I/O from the lock entirely.

DrZlo13 (changes requested):

lib/toolbox/circular_buffer.c
DrZlo13: Maybe we shouldn't reinvent the wheel here and use an existing, well-tested ring buffer library instead?

applications/main/dmesg/dmesg_app.h
Skorpionm: fix

applications/main/dmesg/dmesg_cli.c
Skorpionm: fix

lib/toolbox/circular_buffer.c
Skorpionm: Upon overflow, the tail moves one extra step so that the head does not catch up with the tail.

```
        /* Overwrite: advance tail by exactly (size - space + 1 ) bytes,
           wrapping around the buffer as needed. */
        size_t drop = size - space + 1;
```
But yes, there was an error in determining the amount of free space in the buffer.

applications/main/dmesg/dmesg_cli.c
Skorpionm: I’d rather avoid doing that, as it would require allocating another buffer of the same size for copying. The idea here is to pause the addition of new logs to the dmesg buffer—just as with other logs—while historical data is being output; once all the data has been displayed, normal log processing resumes.

lib/toolbox/circular_buffer.c
Skorpionm: fix

lib/toolbox/circular_buffer.c
Skorpionm: fix

DrZlo13 (approved):

PR #202, opened 2026-07-10
Issue 185 dmesg log history

anon22-byte:
# What's new

- Store recent log output,  then let the user run dmesg to dump it

# Verification 

- I just ran the firmware build again and saw : "ninja: no work to do."

# Contributor checklist


DrZlo13:
Hi. Unfortunately, this isn't quite what we were looking for — we need persistent logging that starts from system boot. Pull request #187 addresses this correctly.

PR #207, opened 2026-07-18
Power: fix BQ28Z620 CLI timing output order

MuratErdimDemir:
## What changed

Fix the BQ28Z620 power CLI `printf` argument order so the following labels display the correct values:

- AvgTimeToEmpty
- TimeToFull
- StandbyCurr
- StandbyTimeToEmpty

No functional behavior is changed; this only corrects the CLI output mapping.

## Verification

- Ran `git diff --check`
- Built the firmware successfully using the existing configured build directory


DrZlo13 (approved):
Nice catch!

PR #212, opened 2026-07-23
CLI: fix screen hard fault when USB disconnects mid-session

### LINKED ISSUE #208: Display emulation in the CLI causes the device to freeze if you disconnect USB while it's running.
If you close the CDC connection while screen is running, the MCU goes into a Hard Fault. While debugging, I traced it down to [exactly this spot](https://github.com/flipperdevices/flipperone-mcu-firmware/blob/dev/applications/services/cli/cli_command_screen.c#L177-L180).

The problem is that the PipeSide* pipe we receive already has its own arrived/broken handlers and is already connected to its own event_loop — but this code overwrites all of that.

I honestly don't know the right way to rewrite this, and we also clearly need furi_check guards against overwriting the callbacks and the event in the pipe — that would have saved me a couple of hours.

munzzyy:
# What's new

Fixes #208 in two parts, plus the guards you asked for.

**Why it hard-faults.** The crash lives in corelibs, not in the overwrite itself. `screen` registers its pipe-broken callback with `FuriEventLoopEventFlagOnce`, and `furi_event_loop_process_event()` frees a `FlagOnce` item during the auto-unsubscribe (it runs before `current_item` is set, so the unsubscribe takes the immediate-free path instead of free-later) and then calls the callback through the freed item. Yanking USB mid-`screen` fires exactly that. I've sent the corelibs fix separately: flipperdevices/fbtng-corelibs#40. On this side, `screen`'s broken callback moves to `0` flags — the same idiom `cli_shell` uses for its own broken callback, and the handler only calls `furi_event_loop_stop()`, which is idempotent. That kills the fault here without waiting on a submodule bump.

**The overwrite you traced.** The shell actually does detach the pipe before spawning a command thread (`cli_shell_execute_command` calls `cli_shell_detach_pipe()` first, and `pipe_attach_to_event_loop()` already `furi_check`s double-attach), but `pipe_detach_from_event_loop()` only cleared `event_loop`. The context and the three callbacks stayed behind. So what `screen` was overwriting were the shell's stale leftovers, silently. Two changes make the handoff contract real:

- `pipe_detach_from_event_loop()` is now the full inverse of attach-plus-wiring: it also clears the callback context and all three callbacks, so a pipe is always handed off pristine.
- The four setters (`context`, `data_arrived`, `space_freed`, `broken`) now `furi_check` that they're not overwriting a live non-NULL value with a different one, the guard you wished had been there. Any future clobbering fails loudly at the call site instead of costing hours in the debugger.

Audited all four wiring users in the tree (`cli_shell.c`, `cli_uart.c`, `cli_vcp.c`, `cli_command_screen.c`): every existing transition is NULL-to-set or set-to-NULL, so nothing trips the new checks. `screen` itself needs no rewiring — with detach doing its job, its attach/set sequence is legitimate ownership transfer, and the shell reinstalls its own wiring after `furi_thread_join()` as before.

# Verification

- Full build with the CI toolchain (arm-none-eabi-gcc 14.2.1, pico-sdk 2.2.0, cmake Release): clean, no new warnings, uf2 produced. Also built once more with the corelibs fix applied to the submodule to confirm the pair composes.
- Call-site audit above for the new `furi_check`s.
- No Flipper One hardware here, so the actual repro needs an on-device pass: run `screen` over CDC, yank USB mid-render (a few times, timing matters), confirm no hard fault and that a reconnect gets a working CLI; then `screen`, Ctrl+C, another command, and `screen` again in the same session to exercise detach/reinstall under the guards. Same subset over the UART CLI wouldn't hurt since it shares the guarded API.

# Contributor checklist

- [x] I have tested these changes myself (on hardware or in simulation) — build-verified and statically traced only; no Flipper One hardware, on-device steps listed above

### DIFF
--- a/applications/services/cli/cli_command_screen.c
+++ b/applications/services/cli/cli_command_screen.c
@@ -177,7 +177,7 @@ static CommandScreen* cli_command_screen_alloc(PipeSide* pipe) {
     pipe_attach_to_event_loop(instance->pipe, instance->event_loop);
     pipe_set_callback_context(instance->pipe, instance);
     pipe_set_data_arrived_callback(instance->pipe, cli_command_screen_click_callback, FuriEventLoopEventFlagEdge);
-    pipe_set_broken_callback(instance->pipe, cli_command_screen_pipe_broken_callback, FuriEventLoopEventFlagOnce);
+    pipe_set_broken_callback(instance->pipe, cli_command_screen_pipe_broken_callback, 0);
 
     furi_event_loop_set_custom_event_callback(instance->event_loop, cli_command_screen_render_callback, instance);
  
--- a/lib/containers/pipe.c
+++ b/lib/containers/pipe.c
@@ -238,10 +238,19 @@ void pipe_detach_from_event_loop(PipeSide* pipe) {
     furi_event_loop_maybe_unsubscribe(pipe->event_loop, pipe_get_instance_semaphore(pipe));
 
     pipe->event_loop = NULL;
+
+    // Detach is the full inverse of attach + wiring up callbacks: clear everything a
+    // new owner would otherwise inherit, so the pipe is handed off in a pristine state.
+    pipe->callback_context = NULL;
+    pipe->on_data_arrived = NULL;
+    pipe->on_space_freed = NULL;
+    pipe->on_pipe_broken = NULL;
 }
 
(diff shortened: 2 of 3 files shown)


DrZlo13 (approved):
Thank you! 

This was a hard problem to solve on our side, we really appreciate it!

PR #215, opened 2026-07-27
cli_uart: remove the leftover double pipe_free in the disable handler

munzzyy:
`CliUartMessageTypeDisable` frees `cli_uart->own_pipe`, sets it to `NULL`, and then a few lines down calls `pipe_free(cli_uart->own_pipe)` again on the now-NULL pointer. `pipe_free()` opens with `furi_check(pipe)`, so that second call crashes the moment the disable path runs.

It looks copy-pasted from `cli_vcp.c`'s disable handler, which frees a second, distinct pipe (`shell_pipe`) at that spot. `cli_uart.c` only has the one pipe. It already freed that pipe above, and leaves the shell side to `cli_shell_free_pipe_on_exit()` the same way `cli_vcp.c` does.

No code calls `cli_uart_disable()` today, so the crash isn't reachable yet. The function is public in `cli_uart.h` though, so the first caller that wires it up would hit it. Dropping the stray call is the whole fix. It builds clean, and there's no caller to exercise it on hardware either way.


Skorpionm:
You don't need to completely trust the AI; removing this free will cause a memory leak.

munzzyy:
The pointer is already NULL when that line runs. Same Disable case, a few lines up (cli_uart.c:173-174):

```c
pipe_detach_from_event_loop(cli_uart->own_pipe);
pipe_free(cli_uart->own_pipe);
cli_uart->own_pipe = NULL;
```

So the call at line 183 is `pipe_free(NULL)`, and `pipe_free()` opens with `furi_check(pipe)` (lib/containers/pipe.c:100), which fires in every build type. The disable path can't leak through that line today because it never gets past it.

The diff hides this, to be fair — the earlier free sits just above the hunk's context window. After the change the pipe is still freed exactly once (line 173), and the shell's side is handled by `cli_shell_free_pipe_on_exit()` from the Enable case, same as before. If you'd rather keep something at that spot I can add a comment noting own_pipe is already gone by then.

DrZlo13 (approved):
Looks good, merging now.

PR #216, opened 2026-07-27
furi_hal_serial: check for NULL before dereferencing in deinit

munzzyy:
`furi_hal_serial_deinit()` reads `serial->periph_ptr` one line before the `if(serial == NULL) return;` guard. That guard is there for exactly the deinit-before-init case the `TODO` right above it calls out, but if `serial` is NULL the dereference already crashed a line earlier.

Moved the NULL check above the `periph_ptr` read. `periph` is only used further down on the non-PIO path, so this is a straight reorder with no behavior change once the handle is initialized. Today's callers all init first, so nothing hits this path yet. It's a public `furi_hal` entry point though, so a later caller could.


DrZlo13 (approved):

DrZlo13:
Thanks, LGTM!

PR #219, opened 2026-07-27
i2c_slave: leave the critical section on the early return

munzzyy:
`i2c_slave_irq_callback()` enters a critical section and then returns out of it without leaving:

```c
static void __isr __not_in_flash_func(i2c_slave_irq_callback)(void) {
    FURI_CRITICAL_ENTER();
    ...
    uint32_t intr_stat = hw->intr_stat;
    if(intr_stat == 0) {
        return;
    }
```

From ISR context `FURI_CRITICAL_ENTER()` resolves to `taskENTER_CRITICAL_FROM_ISR()`, which raises BASEPRI and hands back a mask stored in `__furi_critical_info`. Only `FURI_CRITICAL_EXIT()` passes that mask to `taskEXIT_CRITICAL_FROM_ISR()` and restores it. Taking the early return skips it, so BASEPRI stays raised after the handler finishes and interrupts at or below that priority stop being serviced — the FreeRTOS tick included. Every other exit path in the function does call it.

The `intr_stat == 0` guard is there because the handler can run with nothing latched, which is what makes this reachable: the IRQ fires, the status reads back as 0, and the early return runs. That matters here because this ISR is the CPU↔MCU intercom — `i2c_slave_init()` installs it via `irq_set_exclusive_handler()` at `i2c_slave.c:82`, and `furi_hal_i2c_config.c:253` wires it up as the always-on register interface.

Adding the matching exit before the return builds clean:

```
[1/2] Building C object CMakeFiles/flipperone-mcu-firmware.dir/lib/drivers/i2c_slave/i2c_slave.c.o
[2/2] Linking CXX executable flipperone-mcu-firmware.elf
```

I also scanned both this repo and the pinned corelibs for any other `FURI_CRITICAL_ENTER()` with a `return` before its matching exit. This is the only one.

I don't have hardware, so I haven't watched the stall happen — this comes from the macro expansion and `critical.c`. If you want to confirm on a board, toggling a GPIO inside that `if` and scoping it would show whether the branch is taken in practice.


DrZlo13 (approved):
LGTM!

PR #221, opened 2026-07-27
drv2605l: fail auto-calibration when its status reads fail

munzzyy:
`drv2605l_auto_calibration()` reports success when the I2C reads that decide success fail.

Both reads throw away the return code:

```c
uint8_t go_status = 0;
drv2605l_read_reg(instance, Drv2605lRegGo, &go_status);
...
uint8_t status = 0;
drv2605l_read_reg(instance, Drv2605lRegStatus, &status);
```

`drv2605l_read_reg()` leaves `*data` untouched when the transfer fails, so both locals keep their zero initialiser. Zero happens to be the "everything is fine" value for both: `go_bit == 0` ends the wait loop as though calibration completed, and `diagnostic_result == 0` is no fault. The function then logs `Auto-calibration successful` and returns `true` without the DRV2605L ever confirming it finished, and the calibration data read straight afterwards gets saved off the back of that.

It's on a live path. `haptic_alloc()` calls `haptic_calibration(instance, false)` (`haptic.c:172`), which runs auto-calibration whenever there's no usable NVM calibration data — first boot, after a factory reset, or if that data is corrupt. The forced path from `HapticMessageTypeCalibrate` (`haptic.c:133`) hits it too.

This checks both reads and bails on failure. `if(ret < PICO_OK)` is the same test `drv2605l_get_calibration_data()` already uses in this file for the identical call, so the convention is unchanged, and the two callers in `haptic.c` already handle a `false` return.

Compiles clean:

```
[1/1] Building C object CMakeFiles/flipperone-mcu-firmware.dir/lib/drivers/drv2605l/drv2605l.c.o
```

Same shape as #218 in the IQS7211E driver, where the reads also only populate their output on success.

I don't have hardware, so this comes from reading `drv2605l_read_reg()`'s error paths rather than a bench repro. If you want to confirm it, NACKing the haptic driver mid-calibration should show the current build still logging "successful".

Left alone deliberately: the three `drv2605l_read_reg()` calls below the success log are informational `FURI_LOG_I` lines, so a failure there misreports a logged value but doesn't change the outcome. Happy to fold those in if you'd rather have them consistent.


DrZlo13 (approved):
LGTM

PR #228, opened 2026-07-30
CLI: check malloc result in i2c write/read commands

dmytroyunyk:
noticed while testing the i2c commands that write and read never check
if malloc actually succeeded before writing into the buffer in write
it writes the register byte right after the malloc call in read it
just hands the buffer straight to the i2c driver if malloc fails and
returns null heap gets fragmented pretty easy on this thing your
basically writing reading through address 0 and it hardfaults the
board

fix is small just added a nul check on both allocations if it fails
it prints an error message and bails out same way the other error
cases in this file already do

tested it by building locally and running a few i2c write read
commands on the bus works fine didnt touch anything else in the file

side note first time i tried building the whole thing it failed on
the assets step looked like a race in the python venv setup script
when using parallel two jobs tried creating the same venv at once and
one of them lost ran the build again and it went through clean not
related to my change just something i ran into while testing

### DIFF
--- a/applications/services/cli/cli_command_i2c.c
+++ b/applications/services/cli/cli_command_i2c.c
@@ -136,6 +136,10 @@ static bool cli_command_i2c_write(PipeSide* pipe, FuriString* args) {
         }
 
         data = malloc(data_length + 1); // +1 for device register
+        if (!data) {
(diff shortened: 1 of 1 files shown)


DrZlo13:
Hello.

We never check the result of malloc, since in our system a failed malloc leads to a crash.

PR #231, opened 2026-08-12
A/B partitioning

nminaylov:
# What's new

- Partition table with 2 firmware partitions + 1 data partition for NVM
- NVM takes flash offset/size from partition table

# Verification 

- Copy vscode configs from `vscode_template` folder
- Build and flash via SWD as usual
- If you don't have debug probe, you need to write `partition_table.uf2` first (via mass storage or by `Write partition table` task)
- Check current slot id (A/B) in desktop header

# Contributor checklist


.github/workflows/build.yml
```diff
@@ -244,6 +245,7 @@ jobs:
         # One line per entry of the build matrix
         body: |
           **Compiled firmware for commit `${{needs.firmware.outputs.commit_sha}}`:**
+          - [Partition table](https://update.flipperzero.one/builds/flipper-one-mcu/${{steps.names.outputs.branch_name}}/flipper-one-mcu-partition-table-${{steps.names.outputs.suffix}}.uf2)
```
DrZlo13: empty link, see https://update.flipperzero.one/builds/flipper-one-mcu/nm/ab_partitioning/

targets/f1/pico_flash_region.ld
```diff
@@ -0,0 +1 @@
+FLASH(rx) : ORIGIN = 0x10000000, LENGTH = 0x780000
```
DrZlo13: I don't see how `targets/f1/pico_flash_region.ld‎` is actually used during the build. On top of that, the build generates a `build/pico_flash_region.ld` whose contents don't match this file. This means we've lost overflow protection - the firmware can silently outgrow the FW A partition and the linker won't catch it

Moreover, in MemBrowse `cp "$(find . -name pico_flash_region.ld -print -quit)"` will match both files - the one in `targets/` and the one in `build/` — and which one ends up being copied (and therefore used) is non-deterministic.

scripts/partition_table_update.py
```diff
@@ -0,0 +1,225 @@
...
+
+import argparse
+import json
+import re
+import sys
+from pathlib import Path
+
+DEFAULT_JSON = "targets/f100/partition_table.json"
+DEFAULT_LD = "targets/f100/pico_flash_region.ld"
+DEFAULT_TCL = "targets/f100/flash.tcl"
```
DrZlo13: f100 -> f1

.github/workflows/build.yml
```diff
@@ -185,6 +185,7 @@ jobs:
       if: ${{ !github.event.pull_request.head.repo.fork }}
       run: |
         mv ${{github.workspace}}/firmware/build/flipperone-mcu-firmware.uf2 artifacts/flipper-one-mcu-${{ matrix.target }}-firmware-${SUFFIX}.uf2
+        mv ${{github.workspace}}/firmware/build/partition_table.uf2 artifacts/flipper-one-mcu-partition-table-${SUFFIX}.uf2
```
DrZlo13, line 205: `flipper-one-mcu-partition-table-${SUFFIX}.uf2` doesn't include `matrix.target` in its name, so f1 and f2 produce the same filename and one artifact overwrites the other (with `merge-multiple: true`). Right now the contents happen to be identical, so it's harmless, but the setup is fragile.

I'm not sure what the right fix is here - maybe pull the partition table build out into a separate step, so it's explicitly a single shared build for both targets?

applications/services/desktop/scenes/debug_apps_menu.c
```diff
@@ -76,6 +78,10 @@ static bool debug_menu_input(InputEvent* event, void* context) {
     } else if(event->type == InputTypePress && event->key == InputKey5) {
         furi_hal_power_enter_bootsel();
         consumed = true;
+    } else if(event->type == InputTypePress && event->key == InputKey1) {
+        furi_hal_flash_rollback();
```
DrZlo13, line 83: I might be misunderstanding how it's supposed to work, but I couldn't verify that `furi_hal_flash_rollback` works correctly. My assumption was that by flashing the firmware twice in a row via UF2 I'd fill both slots, and pressing rollback would just free one slot and switch me over to the remaining one. Instead, pressing rollback leaves me with a device that has no firmware at all.

targets/f1/furi_hal/furi_hal_nvm_storage.c
```diff
@@ -0,0 +1,168 @@
...
+}
+
+static int program(blockdevice_t* device, const void* buffer, bd_size_t addr, bd_size_t size) {
+    _safe_flash_update_param_t param = {
+        .is_erase = false,
+        .addr = flash_target_offset(device) + addr,
+        .buffer = (void*)buffer,
+        .size = (size_t)size,
+    };
+    int err = flash_safe_execute(_safe_flash_update, &param, FLASH_SAFE_EXECUTE_TIMEOUT);
```
DrZlo13, line 104: In some places `furi_hal_nvm_storage` use `mutex_enter_blocking` / `mutex_exit`, and in others just `flash_safe_execute`. Is that correct?

vscode_template/tasks.json
```diff
@@ -48,10 +68,10 @@
                 "interface/cmsis-dap.cfg",
                 "-f",
                 "target/${command:raspberry-pi-pico.getTarget}.cfg",
-                "-c",
-                "adapter speed 5000; program \"${command:raspberry-pi-pico.launchTargetPath}\" verify reset exit"
+                "-f",
+                "targets/f1/flash.tcl"
```
DrZlo13: The hardcoded `f1` here - maybe it'd be better to just move `flash.tcl` up a level, into `targets/` directly?

DrZlo13 (changes requested):
Since this PR removes the `.vscode` folder, it'd be worth adding a note in the README about the need to copy the files from `vscode_template` into `.vscode`.

targets/f1/pico_flash_region.ld
nminaylov: This file is generated by [scripts/partition_table_update.py](https://github.com/flipperdevices/flipperone-mcu-firmware/pull/231/files/e932aef611bb311ec3d5db97ddab33eb67c70eab#diff-4e8f4b77254ee8f377904d9238c8de86f5c0b3c957c34a46380552a5189f3955) from partition table, I removed build/pico_flash_region.ld generation from cmake

applications/services/desktop/scenes/debug_apps_menu.c
nminaylov: It's just a debug code, fixed in updater PR

targets/f1/furi_hal/furi_hal_nvm_storage.c
nminaylov: Now it's just a copy of stock blockdevice driver with fixed read offset. Later it will be reworked with furi_hal_flash_* functions, introduced in updater PR

vscode_template/tasks.json
nminaylov: Done

scripts/partition_table_update.py
nminaylov: fixed

.github/workflows/build.yml
nminaylov: fixed

targets/f1/pico_flash_region.ld
nminaylov: fixed

DrZlo13:
Attempting to change settings now results in a crash in
```
void __no_inline_not_in_flash_func(flash_range_program)...
#ifdef PICO_FLASH_SIZE_BYTES
hard_assert(flash_offs + count <= PICO_FLASH_SIZE_BYTES);
#endif
```

DrZlo13 (approved):
Nice work, merging!

PR #240, opened 2026-08-25
Firmware update over I2C

nminaylov:
# What's new

- Updater service to flash MCU firmware from FlipperOS over I2C intercom

# Verification 

- Make sure you have flashed the partition table
- Updating firmware from Flipper OS:
Install required packages:
`sudo apt install gcc-arm-none-eabi cmake python3-periphery python3-venv`
Clone and build MCU firmware (see README.md)
Release I2C bus:
`sudo bash -c "echo 0-0069 > /sys/bus/i2c/drivers/flipper-one-mcu/unbind"`
Run from build dir:
`sudo python3 ../scripts/updater.py --file ./flipperone-mcu-firmware.bin`
Reset MCU to start new firmware

# Contributor checklist


lib/drivers/i2c_slave/i2c_slave.c
```diff
@@ -72,14 +89,15 @@ void i2c_slave_init(i2c_inst_t* i2c, uint8_t address, I2cSlaveCallback callback)
     hw->intr_mask = I2C_IC_INTR_MASK_M_RX_FULL_BITS | I2C_IC_INTR_MASK_M_RD_REQ_BITS | I2C_IC_INTR_MASK_M_TX_ABRT_BITS | I2C_IC_INTR_MASK_M_STOP_DET_BITS |
                     I2C_IC_INTR_MASK_M_START_DET_BITS | I2C_IC_INTR_MASK_M_RESTART_DET_BITS;
```
DrZlo13, line 90: TX_ABRT is still enabled in the interrupt mask but is not handled in the ISR. As a result, the interrupt will keep firing continuously whenever TX_ABRT occurs.

applications/services/updater/updater.c
```diff
@@ -0,0 +1,320 @@
...
+            .type = UpdaterMessageWriteBlockDone,
+            .block.block_idx = updater->i2c_data.idx,
+            .block.block_crc = updater->i2c_data.crc16,
+        };
+        memcpy(msg.block.data, updater->i2c_data.data, sizeof(msg.block.data));
+        furi_check(furi_message_queue_put(updater->message_queue, &msg, 0) == FuriStatusOk);
+    }
+
+    return true;
+}
```
DrZlo13, line 130: `updater->message_queue` holds only two elements, and nothing prevents the host from overflowing this queue except the I2C message rate.

The correct approach here is to return an error rather than crash.

applications/services/i2c_intercom/i2c_intercom.c
```diff
@@ -97,16 +92,19 @@ void __isr __not_in_flash_func(i2c_intercom_isr)(const FuriHalI2cBusHandle* hand
     case FuriHalI2cBusSlaveEventStart:
         // Master has sent a Start signal, prepare to receive address
         instance->state = I2CIntercomStateStart;
-        instance->timeout_alarm = add_alarm_in_ms(I2C_INTERCOM_TIMEOUT_MS, i2c_intercom_timeout_callback, instance, true);
         break;
     case FuriHalI2cBusSlaveEventWrite:
+        if(instance->state == I2CIntercomStateIdle) {
+            furi_crash("I2CIntercom: Write event in invalid state");
+        }
```
DrZlo13, line 99: The correct thing to do here is to reset the i2c state machine rather than crash.

applications/services/updater/updater.c
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+
+static void updater_write_first_sector(Updater* updater) {
+    ImageVersion new_version;
+    bool success = furi_hal_flash_get_image_version(updater->first_sector_buf, new_version.raw);
+    if(!success) {
+        FURI_LOG_E(TAG, "Update image missing version information");
+        return;
+    }
```
DrZlo13, line 148: If the new image has no `version`, the function just logs and returns without changing state. The state remains `UpdaterStateRunning`, and `UpdaterErrorFwVersionError` is declared but never used.

applications/services/updater/updater.c
```diff
@@ -0,0 +1,320 @@
...
+    } else if((msg.type == UpdaterMessageAbortUpdate) && (instance->state != UpdaterStateIdle)) {
+        instance->crc_set = false;
+        instance->size_set = false;
+        instance->fw_blocks = 0;
+        updater_state_change(instance, UpdaterStateIdle, UpdaterErrorNone);
+    } else if((msg.type == UpdaterMessageWriteBlockStart) && (instance->state == UpdaterStateRunning)) {
+        updater_state_change(instance, UpdaterStateBusy, UpdaterErrorNone);
+    } else if((msg.type == UpdaterMessageWriteBlockDone) && (instance->state == UpdaterStateBusy)) {
+        updater_state_change(instance, UpdaterStateRunning, UpdaterErrorNone);
+        updater_process_block(instance, &msg);
```
DrZlo13, line 264: `UpdaterStateRunning` is set before `updater_process_block`, so the host sees `UpdaterStateRunning` while the block is still being written to flash.

This only happens to work because `flash_range_program` runs inside a critical section - the I2C ISR is blocked and the slave stretches SCL. It's probably correct to move `updater_state_change` to after the block is processed.

applications/services/updater/updater.c
```diff
@@ -0,0 +1,320 @@
...
+    if(offset >= sizeof(updater->i2c_data.raw)) return false;
+
+    if(offset == 0) {
+        memset(updater->i2c_data.raw, 0, sizeof(updater->i2c_data.raw));
+        updater->i2c_data_offset = 0;
+        UpdaterMessage msg = {.type = UpdaterMessageWriteBlockStart};
+        furi_check(furi_message_queue_put(updater->message_queue, &msg, 0) == FuriStatusOk);
+    }
+
+    if(offset != updater->i2c_data_offset) return false; // Out-of-order write
```
DrZlo13, line 115: An out-of-order write returns false, but the byte is still ACK'd on the I2C bus, `WriteBlockDone` never arrives, and the state stays `UpdaterStateBusy`. The script spins in `while state == UpdaterStateBusy` forever. 

Both sides need a timeout: the MCU should reset block reception, and the host should bail out of the polling loop.

applications/services/i2c_intercom/i2c_intercom.c
```diff
@@ -97,16 +92,19 @@ void __isr __not_in_flash_func(i2c_intercom_isr)(const FuriHalI2cBusHandle* hand
     case FuriHalI2cBusSlaveEventStart:
         // Master has sent a Start signal, prepare to receive address
         instance->state = I2CIntercomStateStart;
-        instance->timeout_alarm = add_alarm_in_ms(I2C_INTERCOM_TIMEOUT_MS, i2c_intercom_timeout_callback, instance, true);
```
DrZlo13, line 100: There's no recovery mechanism anymore, and `I2C_INTERCOM_TIMEOUT_MS` is now an unused define. Is this change intentional?

targets/f1/furi_hal/furi_hal_flash.c
```diff
@@ -41,17 +79,104 @@ FlashPartitionId furi_hal_flash_get_active_fw_partition(void) {
...
+
+    size_t offset = picobin_ptr - start_ptr;
+    size_t page_nb = offset / FLASH_PAGE_SIZE;
+    size_t page_offset = offset % FLASH_PAGE_SIZE;
+
+    uint8_t* page_buf = malloc(FLASH_PAGE_SIZE);
+    furi_hal_flash_read(part_base + page_nb * FLASH_PAGE_SIZE, page_buf, FLASH_PAGE_SIZE);
+    memset(&page_buf[page_offset], 0, 4);
+    furi_hal_flash_write(part_base + page_nb * FLASH_PAGE_SIZE, page_buf, FLASH_PAGE_SIZE);
+    free(page_buf);
```
DrZlo13, line 110: I dont see an erase before programming and per the W25Q128JV datasheet (Rev F, §8.2.13 Page Program (02h)): 
> The Page Program instruction allows from one byte to 256 bytes (a page) of data to be programmed at previously erased (FFh) memory locations.


So the spec only defines programming for previously erased (FFh) locations; programming over already-programmed data is out-of-spec. In practice on NOR it behaves as an AND (bits can only go 1→0), but the vendor doesn't guarantee it.

To be fair: the same section explicitly allows partial page programming, Winbond doesn't specify a NOP limit per page, and the datasheet has no documented disturb limitations for this part. So overwriting may work in practice - but since we already have the buffer for a data, can we erase the sector and write it properly?

vscode_template/tasks.json
```diff
@@ -32,13 +32,33 @@
...
+                "reveal": "always",
+                "panel": "dedicated"
+            },
+            "problemMatcher": [ ],
+            "windows": {
+                "command": "${env:USERPROFILE}/.pico-sdk/picotool/2.3.1-a4/picotool/picotool.exe"
+            }
+        },
+        {
+            "label": "Write partition table",
```
DrZlo13, line 59: Duplicated task.

lib/drivers/display/display_jd9853_qspi.c
```diff
@@ -240,7 +240,13 @@ static void display_jd9853_qspi_clear(DisplayJd9853QSPI* display) {
...
     DisplayJd9853QSPI* display = (DisplayJd9853QSPI*)ctx;
     furi_hal_gpio_write(&gpio_display_cs, false);
-    display_jd9853_dma_put_buffer(display, (uint8_t*)&display->buffer_header, sizeof(display->buffer_header));
+
+    if(hstx_fifo_hw->stat & HSTX_FIFO_STAT_EMPTY_BITS) {
+        dma_channel_set_read_addr(display->dma_tx_channel, (uint8_t*)&display->buffer_header, false);
+        dma_channel_set_transfer_count(display->dma_tx_channel, sizeof(display->buffer_header) / 4, false);
+        dma_channel_start(display->dma_tx_channel);
+    }
+    // FIXME: what to do if FIFO is not empty?
```
DrZlo13, line 249: There are two `// FIXME` in the code. I understand this is a rework to accommodate the raised I2C priority, but it doesn't look reliable.

Can we resolve the `FIXME` before merging?

targets/f1/furi_hal/furi_hal_flash.c
```diff
@@ -10,6 +10,44 @@
...
+
+static uint8_t* find_picobin_block(const uint8_t* start, size_t size) {
+    for(size_t offset = 0; offset < size; offset++) {
+        const uint8_t* ptr = start + offset;
+        if(*(uint32_t*)ptr == PICOBIN_BLOCK_MARKER_START) {
+            return (uint8_t*)ptr;
+        }
+    }
+    return NULL;
+}
```
DrZlo13, line 23: `find_picobin_block` scans byte-by-byte and casts every offset to `uint32_t*`. Blocks are word-aligned by spec, so use a stride of 4 and bound the loop at size - 4 — right now the last iterations read past the window (`first_sector_buf` in RAM). Also a bare `0xffffded3` match is not enough: the value can appear as a literal in code, so `fw_partition_is_valid` may accept garbage and invalidate may zero code instead of the header. Validate the block like the bootrom does — walk items to `PICOBIN_BLOCK_MARKER_END` within `PICOBIN_MAX_BLOCK_SIZE`.

CMakeLists.txt
```diff
@@ -128,6 +128,7 @@ include(lib/corelibs/lib/version/CMakeLists.txt)
 
 pico_set_program_name(${PROJECT_NAME} "${PROJECT_NAME}")
 pico_set_program_version(${PROJECT_NAME} "0.1")
+pico_set_binary_version(${PROJECT_NAME} MAJOR 0x0010 MINOR 0)
```
DrZlo13, line 131: `MAJOR 0x0010 MINOR 0` is hardcoded, and `MINOR` is rewritten by the updater as a boot counter, so the version in the binary never matches what ends up in flash. `MAJOR` should come from the version lib (`GIT_VERSION` is already computed there), and `MINOR` should be documented as a counter field owned by the updater, not a version.

scripts/updater.py
```diff
@@ -0,0 +1,188 @@
...
+    time.sleep(0.1)  # wait for the last block to be processed
+
+    state, error = get_state(mcu)
+    print(f"State: {state.name}")
+    if state == UpdaterStatus.ERROR:
+        print(f"Error: {error.name}")
+        return
+
+    print(f"Written in {end_time - start_time:.2f}s ({len(firmware_data)/(end_time - start_time)/1024*8:.2f} Kbits/s)")
+    print("Update complete. Reset MCU to start new firmware.")
```
DrZlo13, line 185: After the last block the script sleeps 100 ms and reads the state once. The final CRC over the whole image plus the first-sector write can take longer, so this prints "Update complete" while the MCU is still `RUNNING`. Poll until `DONE` or `ERROR`, with a timeout.

applications/services/updater/updater.c
```diff
@@ -0,0 +1,320 @@
...
+    instance->event_loop = furi_event_loop_alloc();
+    instance->message_queue = furi_message_queue_alloc(2, sizeof(UpdaterMessage));
+    furi_event_loop_subscribe_message_queue(instance->event_loop, instance->message_queue, FuriEventLoopEventIn, updater_message_callback, instance);
+
+    updater_state_change(instance, UpdaterStateIdle, UpdaterErrorNone);
+    instance->active_partition = furi_hal_flash_get_active_fw_partition();
+    furi_hal_flash_get_partition_info(
+        instance->active_partition == FlashPartitionIdFwA ? FlashPartitionIdFwB : FlashPartitionIdFwA,
+        &instance->write_partition_address,
+        &instance->write_partition_size);
```
DrZlo13, line 287: The return value of `furi_hal_flash_get_partition_info` goes unchecked.

scripts/updater.py
```diff
@@ -0,0 +1,188 @@
...
+    ERROR = 4
+
+class UpdaterError(Enum):
+    NONE = 0
+    NOT_STARTED = 1
+    SIZE_ERROR = 2
+    BLOCK_CRC_ERROR = 3
+    BLOCK_INDEX_ERROR = 4
+    FW_CRC_ERROR = 5
+    FW_VERSION_ERROR = 6
```
DrZlo13, line 31: Let's align the naming (`UpdaterStatus` → `UpdaterState`) and add cross-reference "keep in sync" comments in both files, pointing to the counterpart file and type name, e.g. `// Keep in sync with tools/updater.py: UpdaterState`.

DrZlo13 (changes requested):

applications/services/i2c_intercom/i2c_intercom.c
nminaylov: We are I2C slave, we can't do anything if bus hangs. `furi_hal_i2c_slave_bus_reset` breaks state machine

applications/services/i2c_intercom/i2c_intercom.c
nminaylov: This may happen only in case of bug in state machine, reset makes no sense since we are slave

targets/f1/furi_hal/furi_hal_flash.c
nminaylov: All NOR flash support 1->0 overwriting. Also PICOBIN_BLOCK_MARKER_START contains at least 2 0xFF bytes

applications/services/updater/updater.c
nminaylov: When flash writing is started, the i2c part is already ready to start receiving the next block (but can't because of critical section)

lib/drivers/i2c_slave/i2c_slave.c
nminaylov: fixed

applications/services/updater/updater.c
nminaylov: fixed

scripts/updater.py
nminaylov: done

applications/services/updater/updater.c
nminaylov: done

targets/f1/furi_hal/furi_hal_flash.c
nminaylov: done

vscode_template/tasks.json
nminaylov: done

scripts/updater.py
nminaylov: This script is a simple POC. I'm not sure it should be made more complex

applications/services/updater/updater.c
nminaylov: We can't do anything from MCU side, only remain in busy state until full block write (or until Abort command)

applications/services/updater/updater.c
nminaylov: Returning error makes no sense - we can't respond NAK

CMakeLists.txt
nminaylov: GIT_VERSION don't contain valid semver on branches. We need to store previous release version somewhere or bump this value manually

PR #243, opened 2026-09-01
Desktop: start/stop app from CLI

Skorpionm:
# What's new

- Desktop: start/stop app from CLI
- Furi_bsp_expander: reporting a false memory leak error
- DRV2605L: fix reboot

# Verification 

- launch the MCU CLI
- use the commands "desktop start_app <app_name|appid>", "desktop stop_app", and "desktop list_apps", and "get_run_app"

image

# Contributor checklist

### DIFF
--- a/applications/services/desktop/desktop.c
+++ b/applications/services/desktop/desktop.c
@@ -18,13 +18,18 @@
 
 typedef enum {
     DesktopMessageTypeAppStart,
+    DesktopMessageTypeAppStop,
     DesktopMessageTypeAppClosed,
+    DesktopMessageTypeAppRegister,
+    DesktopMessageTypeAppUnregister,
 } DesktopMessageType;
 
 typedef struct {
     DesktopMessageType type;
     const FlipperInternalApplication* app;
     const char* args;
+    const char* appid;
+    FuriThread* thread;
 } DesktopAppMessage;
 
 typedef struct {
@@ -34,8 +39,10 @@ typedef struct {
 
 typedef struct {
     bool running;
+    bool external;
     char* args;
     FuriThread* thread;
+    const char* name;
 } DesktopApp;
 
 struct Desktop {
@@ -112,6 +119,7 @@ static void desktop_do_app_closed(Desktop* desktop) {
 
     furi_thread_free(desktop->app.thread);
     desktop->app.thread = NULL;
+    desktop->app.name = NULL;
 
     FURI_LOG_I(TAG, "Application stopped. Free heap: %zu", memmgr_get_free_heap());
 }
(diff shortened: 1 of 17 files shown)


applications/services/desktop/desktop.c
```diff
@@ -293,28 +415,116 @@ bool desktop_start_app(const FlipperInternalApplication* app) {
...
+
+    DesktopAppMessage message = {
+        .type = DesktopMessageTypeAppUnregister,
+        .appid = appid,
+    };
+    furi_message_queue_put(desktop->app_message_queue, &message, FuriWaitForever);
+    furi_record_close(RECORD_DESKTOP);
+
+    return true;
+}
```
DrZlo13, line 597: TOC/TOU here. 

If `desktop_register_app` and `desktop_unregister_app` happen to be called within the same thread and placing their results into the queue before the desktop processes it - the desktop's queue will end up containing only `DesktopMessageTypeAppRegister`, causing the desktop to get stuck permanently in the `already running` state.

The correct approach here is to make the code synchronous using an `api_lock` or even better, the desktop itself should handle the autostart of applications.

applications/services/desktop/desktop.c
```diff
@@ -112,10 +120,48 @@ static void desktop_do_app_closed(Desktop* desktop) {
...
+
+    if(cleanup->args) {
+        free(cleanup->args);
+    }
+
+    furi_thread_free(cleanup->thread);
+    free(cleanup);
+
+    FURI_LOG_I(TAG, "Application stopped. Free heap: %zu", memmgr_get_free_heap());
+}
```
DrZlo13: Calling `furi_thread_join` from the timer thread will halt all system timers until the corresponding thread terminates.
It is better to perform this from the desktop thread.

applications/services/desktop/desktop_cli.c
```diff
@@ -0,0 +1,118 @@
...
+    bool handled = false;
+
+    if(args_read_string_and_trim(args, cmd)) {
+        const char* cmd_str = furi_string_get_cstr(cmd);
+        for(size_t i = 0; i < COUNT_OF(desktop_cmds); i++) {
+            const DesktopCmd* c = &desktop_cmds[i];
+            if(strcmp(cmd_str, c->name) == 0) {
+                if(!c->execute(pipe, args)) {
+                    printf("usage: desktop %s %s\r\n", c->name, c->arg_spec);
+                }
```
DrZlo13, line 124: All `desktop` commands return `true`, so this code will never execute.

It is better to return `false` instead of `printf("usage: desktop start_app...`

DrZlo13 (changes requested):

DrZlo13 (approved):

PR #247, opened 2026-09-09
Fix pixel shift

Skorpionm:
# What's new

- Fix pixel shift

# Verification 

- Launch Linux and press the "Power" button several times; the image should not shift.

# Contributor checklist

### DIFF
--- a/lib/drivers/pio_get_frame/pio_get_frame.c
+++ b/lib/drivers/pio_get_frame/pio_get_frame.c
@@ -124,15 +124,6 @@ PioGetFrame* pio_get_frame_init(const GpioPin* gpio_cs, const GpioPin* gpio_sck,
     instance->callback_rx = NULL;
     instance->callback_context = NULL;
 
-    /* The DMA/drain write path never touches data[0] (it starts at
-     * +PIXEL_SHIFT_BUG_WORKAROUND), so it must be cleared explicitly instead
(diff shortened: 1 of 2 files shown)


lib/drivers/pio_get_frame/pio_get_frame.c
```diff
@@ -115,6 +124,15 @@ PioGetFrame* pio_get_frame_init(const GpioPin* gpio_cs, const GpioPin* gpio_sck,
...
 
+    /* The DMA/drain write path never touches data[0] (it starts at
+     * +PIXEL_SHIFT_BUG_WORKAROUND), so it must be cleared explicitly instead
+     * of being left with whatever malloc() handed us. */
+    for(size_t i = 0; i < PIO_GET_FRAME_COUNT; i++) {
+        for(size_t j = 0; j < PIO_GET_FRAME_PIXEL_SHIFT_BUG_WORKAROUND; j++) {
+            instance->frame_buffers[i].data[j] = 0;
+        }
+    }
+
```
DrZlo13: It is better to use `memset` here, but generally, even `memset` isn't needed: our `malloc` already returns zero-initialized memory.

DrZlo13 (changes requested):

DrZlo13 (approved):
LGTM, thank you!

PR #248, opened 2026-09-10
Fixed the mechanism for entering MaskROM mode.

Skorpionm:
# What's new

- Cpu_app: Fixed the mechanism for entering MaskROM mode

# Verification 

# Contributor checklist

### DIFF
--- a/applications/main/cpu/cpu_app.c
+++ b/applications/main/cpu/cpu_app.c
@@ -16,6 +16,7 @@
 #define TAG "CpuApp"
 
 #define CPU_APP_MESSAGE_QUEUE_SIZE 64
+#define CPU_APP_MASKROM_TIMER_INTERVAL 1000
 
 #define CPU_ARG_START   "start"
 #define CPU_ARG_MASKROM "maskrom"
@@ -56,6 +57,7 @@ typedef struct {
     FuriMessageQueue* app_queue;
     PioGetFrame* pio_get_frame;
     size_t skip_frames;
+    FuriEventLoopTimer* maskrom_timer_off;
 } CpuApp;
 
 static void furi_hal_reset_pd_and_charger(void) {
@@ -167,6 +169,13 @@ static void __isr __not_in_flash_func(cpu_app_pio_get_frame_isr)(uint8_t* data, 
     }
 }
 
+static void cpu_app_timer_callback(void* context) {
+    furi_assert(context);
+    UNUSED(context);
+
+    furi_bsp_linux_maskrom(false);
+}
+
 static void cpu_app_message_logic(FuriEventLoopObject* object, void* context) {
     furi_check(context);
     CpuApp* instance = context;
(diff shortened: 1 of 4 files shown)


applications/services/power/power_cli.c
```diff
@@ -226,7 +233,8 @@ static const PowerCmd power_cmds[] = {
     {"ship", "", "Enter ship mode", power_cli_ship_mode},
     {"reboot", "", "Reboot the device", power_cli_reboot},
     {"boot", "", "Boot MCU to bootloader", power_cli_boot},
-    {"maskrom", "", "Boot CPU to maskrom", power_cli_maskrom},
+    {"maskrom_on", "", "Boot CPU to maskrom", power_cli_maskrom_on},
+    {"maskrom_off", "", "Exit maskrom mode", power_cli_maskrom_off},
```
DrZlo13, line 237: `maskrom on/off`?

I also don't like that the behaviour differs between `cpu_app` and the CLI: in one place we wait and then release the pin, in the other we don't.

Can we make them consistent? Or if this is intentional, let's state it explicitly?

applications/main/cpu/cpu_app.c
```diff
@@ -167,6 +169,13 @@ static void __isr __not_in_flash_func(cpu_app_pio_get_frame_isr)(uint8_t* data,
     }
 }
 
+static void cpu_app_timer_callback(void* context) {
+    furi_assert(context);
+    UNUSED(context);
```
DrZlo13, line 174: `furi_assert(context)` immediately followed by `UNUSED(context)`.

applications/main/cpu/cpu_app.c
```diff
@@ -284,8 +296,13 @@ static void cpu_app_free(CpuApp* instance) {
     view_free(instance->menu_view);
 
     furi_event_loop_unsubscribe(instance->event_loop, instance->app_queue);
+    if(furi_event_loop_timer_is_running(instance->maskrom_timer_off)) {
+        furi_event_loop_timer_stop(instance->maskrom_timer_off);
+    }
+    furi_event_loop_timer_free(instance->maskrom_timer_off);
```
DrZlo13, line 302: `furi_event_loop_timer_free` alone is enough here - it already stops the timer.

DrZlo13 (changes requested):

applications/main/cpu/cpu_app.c
Skorpionm: Definitely not here! What if the user opens and immediately closes the app? The timer would still be running, and we’d end up deleting an active timer.

applications/services/power/power_cli.c
Skorpionm: This logic has been completely removed; launching maskrom mode will be handled via the cpu_app_maskrom application.

DrZlo13 (approved):
Thank you, merging.

PR #250, opened 2026-09-11
Cpu_mode srv

Skorpionm:
# What's new

- Add Cpu_mode srv
- Automatic shutdown of Cpu_app when Linux shuts down via the "shutdown -h now" command.

# Verification 

# Contributor checklist

### DIFF
--- a/applications/services/cpu_mode/cpu_mode.c
+++ b/applications/services/cpu_mode/cpu_mode.c
@@ -1,6 +1,7 @@
 #include "cpu_mode.h"
 
 #include <api_lock.h>
+#include <desktop/desktop.h>
 
 #define TAG "CpuMode"
 
@@ -17,8 +18,8 @@ struct CpuMode {
 };
 
 typedef enum {
-    CpuModeMessageTypeSetCpuMode,
-    CpuModeMessageTypeGetCpuMode,
+    CpuModeMessageTypeSetCpuState,
+    CpuModeMessageTypeGetCpuState,
 } CpuModeMessageType;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


applications/services/cpu_mode/cpu_mode.c
```diff
@@ -0,0 +1,168 @@
...
+    CpuModeMessage msg;
+    furi_check(furi_message_queue_get(instance->message_queue, &msg, 0) == FuriStatusOk);
+
+    bool result = false;
+
+    switch(msg.type) {
+    case CpuModeMessageTypeSetCpuState: {
+        CpuState cpu_state = *msg.as.cpu_state.cpu_state;
+        if(cpu_state >= CpuStateNumStates) {
+            // Ignore garbage from the I2C link (bus glitch, protocol/version
```
DrZlo13, line 52: This comment cuts off mid-sentence.

applications/services/cpu_mode/cpu_mode.c
```diff
@@ -0,0 +1,168 @@
...
+        .type = CpuModeMessageTypeSetCpuState,
+        .result = &result,
+        .lock = api_lock_alloc_locked(),
+        .as =
+            {
+                .cpu_state =
+                    {
+                        .cpu_state = &cpu_state,
+                    },
+            },
```
DrZlo13, line 143: We can pass `cpu_state` here directly, by value rather than through a pointer. Wrapping it in a `cpu_state` struct also seems unnecessary. As a bonus, this frees us from having to wait the `api_lock`.

applications/services/cpu_mode/cpu_mode.c
```diff
@@ -0,0 +1,168 @@
...
+
+static void cpu_mode_shutdown_timer_callback(void* context) {
+    furi_check(context);
+    CpuMode* instance = context;
+    if(instance->status.cpu_state == CpuStatePoweredOff) {
+        const char* appid = desktop_get_running_app_id();
+        if(appid && (!strcmp(appid, "cpu_app_start") || !strcmp(appid, "cpu_app_maskrom"))) {
+            FURI_LOG_D(TAG, "Cpu_app stopping due to CPU State: %d", instance->status.cpu_state);
+            desktop_stop_app();
+        }
```
DrZlo13, line 103: There's a race between `desktop_get_running_app_id()` and `desktop_stop_app()`: if `cpu_app` exits and the user launches another app in between, we'll stop the wrong one. A `desktop_stop_app_if(app_id)` (or maybe `desktop_stop_app_by_id(app_id)`) that does the compare-and-stop inside the `desktop` thread as a single message would close the race.

DrZlo13 (changes requested):

applications/services/cpu_mode/cpu_mode.c
Skorpionm: Changed to pass-by-value; kept the locking, since the result function needs to execute.

DrZlo13 (changes requested):
Sorry for not bringing this up right away, the thought only just occurred to me.

A whole thread with an event loop, a message queue and an api_lock is created for a single enum with a single writer. I think we can do the same thing with a `FuriState` (like `led_get_brightness_state`) or a `FuriPubSub` (like `headphones`), and put the timer on the existing event loop in `cpu_app`. Then `cpu_app` subscribes to the state itself and triggers its own `CpuAppMessageTypeShutdown`, so the hardcoded `"cpu_app_start"/"cpu_app_maskrom"` strings in `cpu_mode.c` and the desktop changes become unnecessary. This significantly reduces the amount of code, entities and API surface.

PR #253, opened 2026-09-16
Cli: Add deletion of the character to the right when the Delete key is pressed.

### LINKED ISSUE #252: CLI line edit doesn't know DEL (only Backspace)
MCU CLI interface supports command line editing, but trying to "erase forward" with the Delete key (as opposed to "erase backward" with the Backspace key) doesn't remove characters but instead inserts garbage into the command line

Skorpionm:
# What's new

- Cli: Add deletion of the character to the right when the Delete key is pressed.
- Cli: Fix а handler for F1..F5 keys—which use sequences like ESC[1X~—was previously being intercepted by the Backspace parser.

# Verification 

- Launch the CLI and verify the handling of function keys, Backspace, Delete, Home, and others.

# Contributor checklist


UnsleptArch:
**What the bug actually traces to**

Hello I hope you guys are doing great, the reported symptom is that pressing Delete inserts a garbage character into the command line. I spent some time tracing this through the actual parser in lib/cli/cli_ansi.c, and the problem seems to come down to how the ANSI escape sequence is currently being handled.

Most terminals, including Konsole and other terminals following the usual xterm conventions, send the Delete key as the four-byte sequence ESC [ 3 ~ (0x1B 0x5B 0x33 0x7E). This is different from the raw 0x7F byte (CliKeyDEL), which is what many terminals send for the physical Backspace key instead. So, in other words, these are two separate keys producing two separate inputs, while the current parser only really handles the latter. (Xterm conventions can be just a little confusing sometimes.)

I walked through the parser state machine one byte at a time with the actual sequence, and this is what happens:

0x1B, which is ESC, is received, so the parser changes to the Escape state.
0x5B, the [ character, is received, so the parser changes to the EscapeBrace state.
0x33, the digit 3, is received. In EscapeBrace, the code currently has a special case for the digit 1 (used for modified arrows such as Ctrl+Left), but otherwise treats the byte as a key mnemonic. cli_ansi_key_from_mnemonic('3') does not match any of its supported cases (A, B, C, D, F, or H), so it returns Unrecognized. At that point, the parser resets instead of waiting for the final byte of the sequence.
0x7E, the ~ character and final byte of the Delete sequence, then arrives while the parser is back in the Initial state. Since it is not an ESC, it is returned immediately as a literal keystroke.

As a result, the ~ gets inserted into the command line.

So the actual behavior is essentially one stray ~ character for every Delete keypress, while Delete itself does not perform any useful action.

I also verified this independently by extracting cli_ansi.c and cli_ansi.h into a small host-side test harness, using the real parser files, a minimal furi.h stub, and GCC. I then fed the literal four-byte sequence into the parser:

```
Feeding Delete key sequence (ESC [ 3 ~) byte by byte:
  byte 0x1B -> is_done=0
  byte 0x5B -> is_done=0
  byte 0x33 -> is_done=1, key=Unrecognized
  byte 0x7E -> is_done=1, key='~'
```

That reproduces the reported behavior directly at the parser level.

**Why pull request 253 does not fully close this**

Pull request 253 addresses a related part of the problem by reassigning CliKeyDEL, the raw 0x7F byte, from backspace to forward delete. However, this does not change the ANSI escape sequence parser itself.

For a terminal that sends ESC[3~ for Delete, which is a very common convention, the parser behavior described above still occurs. In that case, the fix from PR 253 never actually reaches the problematic input sequence.

There is also a potential regression for terminals or configurations where the physical Backspace key sends the raw 0x7F byte, which is a real and fairly common convention. Those users would lose the existing backward-delete behavior without gaining a working forward-delete key, because their physical Delete key would still be sending ESC[3~, not 0x7F. However I mean that in no shade to the wonder of that PR, and I think that they're solution was super close to fully closing this, as a fellow cyber deck / flipper enthusiast I think them for the work they done.

**The fix**

The proposed change adds an explicit parser state for ESC[3~ and resolves that sequence to a new CliKeyDelete.

The existing CliKeyDEL handling and Backspace binding are left unchanged, so the new handling is isolated to the actual Delete escape sequence rather than repurposing the existing Backspace input.

```diff
diff --git a/lib/cli/cli_ansi.c b/lib/cli/cli_ansi.c
index 816758b..d3347d3 100644
--- a/lib/cli/cli_ansi.c
+++ b/lib/cli/cli_ansi.c
@@ -7,6 +7,9 @@ typedef enum {
     CliAnsiParserStateEscapeBraceOne,
     CliAnsiParserStateEscapeBraceOneSemicolon,
     CliAnsiParserStateEscapeBraceOneSemicolonModifiers,
+    // <ESC> [ 3 ..., waiting on the trailing '~' of the Delete key's
+    // sequence (ESC [ 3 ~).
+    CliAnsiParserStateEscapeBraceDelete,
 } CliAnsiParserState;
 
 struct CliAnsiParser {
@@ -81,12 +84,27 @@ CliAnsiParserResult cli_ansi_parser_feed(CliAnsiParser* parser, char c) {
         break;
 
     case CliAnsiParserStateEscapeBrace:
-        // <ESC> [ <key mnemonic> -> <key>
-        if(c != '1') PARSER_RESET_AND_RETURN(parser, CliModKeyNo, cli_ansi_key_from_mnemonic(c));
-
         // <ESC> [ 1 ...
-        parser->state = CliAnsiParserStateEscapeBraceOne;
-        break;
+        if(c == '1') {
+            parser->state = CliAnsiParserStateEscapeBraceOne;
+            break;
+        }
+
+        // <ESC> [ 3 ... (Delete key, expects a trailing '~')
+        if(c == '3') {
+            parser->state = CliAnsiParserStateEscapeBraceDelete;
+            break;
+        }
+
+        // <ESC> [ <key mnemonic> -> <key>
+        PARSER_RESET_AND_RETURN(parser, CliModKeyNo, cli_ansi_key_from_mnemonic(c));
+
+    case CliAnsiParserStateEscapeBraceDelete:
+        // <ESC> [ 3 ~ -> Delete. Consume the trailing byte either way so an
+        // unexpected one doesn't leak back into the input stream as a
+        // literal character the way it used to.
+        if(c == '~') PARSER_RESET_AND_RETURN(parser, CliModKeyNo, CliKeyDelete);
+        PARSER_RESET_AND_RETURN(parser, CliModKeyNo, CliKeyUnrecognized);
 
     case CliAnsiParserStateEscapeBraceOne:
         // <ESC> [ 1 <non-;> -> error
diff --git a/lib/cli/cli_ansi.h b/lib/cli/cli_ansi.h
index 6f0a2f8..1f841df 100644
--- a/lib/cli/cli_ansi.h
+++ b/lib/cli/cli_ansi.h
@@ -102,6 +102,10 @@ typedef enum FURI_PACKED {
     CliKeyDown,
     CliKeyHome,
     CliKeyEnd,
+    // Forward-delete, sent by real terminals as ESC [ 3 ~ — distinct from
+    // CliKeyDEL (raw 0x7F), which most terminals send for the physical
+    // Backspace key instead.
+    CliKeyDelete,
 } CliKey;
 static_assert(sizeof(CliKey) == sizeof(char));
 
diff --git a/lib/cli/shell/cli_shell_line.c b/lib/cli/shell/cli_shell_line.c
index 0de6f33..8da38f7 100644
--- a/lib/cli/shell/cli_shell_line.c
+++ b/lib/cli/shell/cli_shell_line.c
@@ -325,6 +325,33 @@ static bool cli_shell_line_input_bksp(CliKeyCombo combo, void* context) {
     return true;
 }
 
+static bool cli_shell_line_input_del(CliKeyCombo combo, void* context) {
+    UNUSED(combo);
+    CliShellLine* line = context;
+    // erase one character forward, at the cursor, without moving it — the
+    // mirror image of bksp above, which erases behind the cursor and moves
+    // it back by one.
+    cli_shell_line_ensure_not_overwriting_history(line);
+    FuriString* editing_line = cli_shell_line_get_editing(line);
+    cli_shell_line_clamp_position(line, editing_line);
+    if(line->line_position >= furi_string_size(editing_line)) {
+        putchar(CliKeyBell);
+        stdio_flush();
+        return true;
+    }
+    furi_string_replace_at(editing_line, line->line_position, 1, "");
+
+    // print the rest of the line, restore cursor (no initial left-shift:
+    // the cursor doesn't move for a forward delete)
+    printf(
+        "%s" ANSI_ERASE_LINE(ANSI_ERASE_FROM_CURSOR_TO_END),
+        furi_string_get_cstr(editing_line) + line->line_position);
+    size_t left_by = furi_string_size(editing_line) - line->line_position;
+    if(left_by) printf(ANSI_CURSOR_LEFT_BY("%zu"), left_by);
+    stdio_flush();
+    return true;
+}
+
 static bool cli_shell_line_input_ctrl_l(CliKeyCombo combo, void* context) {
     UNUSED(combo);
     CliShellLine* line = context;
@@ -399,7 +426,7 @@ static bool cli_shell_line_input_normal(CliKeyCombo combo, void* context) {
 
 CliShellKeyComboSet cli_shell_line_key_combo_set = {
     .fallback = cli_shell_line_input_normal,
-    .count = 14,
+    .count = 15,
     .records =
         {
             {{CliModKeyNo, CliKeyETX}, cli_shell_line_input_ctrl_c},
@@ -413,6 +440,7 @@ CliShellKeyComboSet cli_shell_line_key_combo_set = {
             {{CliModKeyNo, CliKeyEnd}, cli_shell_line_input_end},
             {{CliModKeyNo, CliKeyBackspace}, cli_shell_line_input_bksp},
             {{CliModKeyNo, CliKeyDEL}, cli_shell_line_input_bksp},
+            {{CliModKeyNo, CliKeyDelete}, cli_shell_line_input_del},
             {{CliModKeyNo, CliKeyFF}, cli_shell_line_input_ctrl_l},
             {{CliModKeyCtrl, CliKeyLeft}, cli_shell_line_input_ctrl_left_right},
             {{CliModKeyCtrl, CliKeyRight}, cli_shell_line_input_ctrl_left_right},
```

**Verification**

I ran the same host-side harness again after applying the patch, this time with a few additional regression checks.

The Delete sequence now correctly produces:

```
ESC [ 3 ~ -> key=Delete
```

with no leftover ~ byte being passed through to the command line.

I also checked that:

The Up Arrow sequence still produces key=Up.
The raw Backspace byte 0x7F still produces key=DEL.
Left Arrow with modifiers still produces key=Left, mods=4, confirming that the existing digit-1 prefixed escape-sequence handling shared by the new branch is still working as expected.

I don't have the hardware available to flash and test this on the actual device (duh), so I want to be clear about that rather than claiming hardware verification that I haven't performed. The fix has been verified on a host machine against the actual parser logic, but it has not yet been verified on physical hardware. I'm hoping that you guys will be able to perform that to make sure that there's no slippery edge case that I missed.

DrZlo13 (changes requested):
On my system (macOS, zsh), pressing Delete just inserts a `~` character into the MCU console.

DrZlo13:
@UnsleptArch
> I don't have the hardware available to flash and test this on the actual device (duh), so I want to be clear about that rather than claiming hardware verification that I haven't performed.

Our firmware should run on a standard Pico2 board - at least, that was the case about a month ago. This should be sufficient to test the console's behavior.

Besides, your point makes sense, @Skorpionm - take a closer look please.

Skorpionm:
In my system (Windows 10), the backspace and delete keys send exactly 1 byte each (like in PR). I checked it on teraterm, puty.

UnsleptArch:
Thanks for the reply. Just to answer that directly, I don't have a Pico2 or any Flipper One hardware myself, and unfortunately I can't get one right now :(, so I'm not able to flash test this on the actual hardware even though the firmware apparently does run on stock Pico2 hardware which is pretty neat btw.

I just went ahead and wrote up the parser level fix I described above as a full PR rather than leaving it as just a comment, mostly in case it's useful to build on directly instead of having to rederive the fix from the explanation. It's #258.

I've verified it with a standalone host side test of the parser logic, feeding the actual byte sequences in and checking the resulting CliKeyCombo output, including regression checks for the existing arrow key paths. I just haven't been able to verify it with real hardware or actual terminal I/O.

@Skorpionm or anyone else who wants to, feel free to take it, adapt it, or fold it into #253 however makes the most sense. I mainly just wanted to have the actual fix available so there's something concrete to work from rather than only describing it in prose and causing any confusion.

UnsleptArch:
I went a bit further into the whole “one byte for both keys” point, and it turns out this is actually a real terminal compatibility issue rather than just a simple conf mistake.
Tera Term for example, has a Delete key option that can send a raw 0x7F for Delete instead of ESC[3~. PuTTY's Delete behavior can also vary depending on the terminal type and keypad mode. Interestingly, xterm itself used 0x7F for Delete historically, up until 1998 (I didn't even know it went that far back, you learn something new everyday ig).
So the distinction here is more nuanced than i initially made it sound.
With this patch, 0x7F continues to mean Backspace and ESC[3~ maps to Delete. That's the behavior that makes sense for the xterm and Linux terminal defaults this parser is primarily trying to handle. However, there are definitely serial terminal configurations where the actual Delete key sends a bare 0x7F, with no way to distinguish it from Backspace at the byte level.
That means #253 covers the terminals that use raw 0x7F for Delete, while this PR covers terminals that use the standard ESC[3~ sequence. Unfortunately, a bare 0x7F doesn't contain enough information to determine which physical key generated it, so a single parser mapping can't correctly support both interpretations at the same time without some kind of configuration or terminal mode selection.
I wanted to just mention that because I don't want to leave the earlier “most terminals” statement as if that's the whole story. There are clearly valid setups on both sides of this, and I think it's worth making that compatibility tradeoff explicit.
I'm open to whatever approach makes the most sense here. If a maintainer would rather keep the current behavior, add a configuration option, or combine the two approaches in some other way, I'm happy to adapt the PR. I mainly wanted to make sure this edge case was documented now that I've dug into the actual terminal behavior a little bit more.
Best regards,
Unslept

Skorpionm:
@UnsleptArch 
Forgive me, but stop generating tons of AI text. Honestly, it's not interesting to read and disgusting. I know how to use it too. Take a close look at the current PR and test it on real hardware. The current implementation handles everything correctly, even the F1, F2, Fx keys... and not just delete and backspace. Write your thoughts in your own words.

UnsleptArch:
damn man its fine but i just cant push out as many quality PR's hand writing everything and the quality would sink, only reason i'm even spending my time on this is because the flipper community asked for help, just wanted to contribute what i could sorry man ill drop my prs

UnsleptArch:
If the contributing guidelines had outlined that, I wouldn't have used LLMs for writing either. I have nothing against that, and besides the hardware, which I don't have and can't get, I fully filled out the checklist and got everything you guys wanted done too. It is one thing if I had used AI when I was told not to, but it literally says, "If you used AI tools to assist with this PR, that is fine, but the checklist above still applies to you personally." I don't know what I did to even provoke you guys. Good luck on this project in the future.

UnsleptArch:
Like you realize that no one in the community gains anything from helping you guys besides potentially getting this project (which probably wont ever even come out,) maybe slightly faster out? If you guys don't want LLM produced PR's, then explicitly say do, which actually you guys said LLM's are fine as long as the rest of the checklist is filled out. You think that I'm gaining anything from helping at all, I'm not, I'm sinking time instead into a project that pays your guy's paycheck that you guys asked help for, in a way that YOUR contribution guidelines explicitly allowed. Just think about that for a second. I'm sorry if i came out hotheaded here but I could have potentially just gotten a "Please write your PR's in your own words" notice beforehand, which i would have been more then happy to oblige with, but again the LLM writes better then me, and I save my time, while getting a clearer point across, so in there besides you not liking LLM written PR's what exactly is the problem? Also, on "the current implementation handles everything correctly, even the F1, F2, Fx keys" I check your diff and there's no case on F5-F12, which btw use the exact  ESC[<n>~ form and F1–F4 don't even reach this function since they're sent as ESC O P/Q/R/, which is a completely different sequence all together.

Skorpionm:
I have nothing against LLM, and yes, it's convenient to use and often speeds up development. But it needs to be fully monitored and the resulting code should be checked, not just generated "so many high-quality PRs." I'm talking specifically about people simply making a request to LLM and mindlessly uploading the resulting code to GIT (I'm not talking about you, but in general). As for the sequences that "aren't processed," I don't have any, since we don't need them now, and the generated code needs to be maintained. But the parser discards function key presses instead, not allowing junk to reach the console.

UnsleptArch:
Sure ig man, for reference all my PR's (closed now as per what you most likely wanted) were monitored the code WAS checked and verified to the best of my extent and were fixing the exact issues you guys wanted help with. For future note the earlier "Your ai text is disgusting" message you sent earlier claimed 'The current implementation handles everything correctly, even the F1, F2, Fx keys... and not just delete and backspace. ' turned into 'those sequences aren't processed, we don't need them' when I pointed that out with your diff, might want to fix that.

UnsleptArch:
@Skorpionm Looking back at this whole event, I realize that I was probably in the wrong, and I would like to apologize for getting hot headed, however I hope you can understand where I'm coming from and how I got frustrated, any comments I could have made regarding the overall state of this project and the nature of my contributions were totally unnecessary, I'm sorry. Technically speaking I think that this patch is fine, however I think that there is a real distinction between terminals that send 0x7F for Delete and those that send ESC[3~, and that's not really a hard gap to cover, however that's more off being nit-picky and isn't totally required for this process, I also think that the F1, F2, Fx keys... are easy to add, and could be worth doing, just so it doesn't come back harshly in the future.

I also understand your point on LLM's and why you might have gotten frustrated yourself, LLM's feel extremely soulless and lazy, especially in the world of PR's, I used them for my contributions on the project because they're simply better then me at writing code and writing writing, but yea either-ways I'm going to take this experience as a lesson and try to write my future PR's, not just on this project but others too, by hand.

If you want me to, I can go back and rewrite my prior patches in my own words / code and resubmit them, or help out with anything else you guys could want in this project, If not I understand and best of luck to you.

Sincerely,
Unslept.

UnsleptArch:
Mb I hadn't realized you had changed the patch since my last look, so my technical claims aren't all valid anymore, I think this patch is complete now besides xterm-style F1–F4 (ESC O P) and certain sequences like ESC[3;5~ leaking a stray character but it was like that before this patch as well.

